PR #157, opened 2014-12-16
Make database validation optional

robinbowes:
We don't want to validate the DB connection when building puppetdb nodes as we're building AMIs on AWS that won't necessarily have access to the RDS instance they will eventually use when instantiated.

This PR adds a couple of flags to make the db validation optional, one for database, one for read_database.

R.


manifests/server/read_database_ini.pp
```diff
@@ -42,12 +45,16 @@
...
+    $ini_setting_require = str2bool($database_validate) ? {
+      false => undef,
+      default  => Class['puppetdb::server::validate_read_db'],
+    }
     # Set the defaults
     Ini_setting {
       path    => "${confdir}/read_database.ini",
       ensure  => present,
       section => 'read-database',
-      require => Class['puppetdb::server::validate_db'],
```
wkalt: we're replacing this with puppetdb::server::validate_read_db -- was that an existing bug or is this changing behavior? ping @kbarber

kbarber:
@robinbowes ping :-).

manifests/server/read_database_ini.pp
robinbowes: This is a change of behaviour. By all means switch to validate_read_db as necessary.

My only requirement is that I can disable validation so we can build AWS AMIs without requiring a postgresql connection.

kbarber: From the top of my head I think the change to validate_read_db is correct IMHO. If it works, I would keep it and call the old way a bug.

kbarber:
@robinbowes see previous comments, also I think this needs a rebase now.

robinbowes:
Rebased.

manifests/server/database_ini.pp
```diff
@@ -18,29 +19,35 @@
...
+      database_port     => $database_port,
+      database_username => $database_username,
+      database_password => $database_password,
+      database_name     => $database_name,
+    }
   }
 
+  $ini_setting_require = str2bool($database_validate) ? {
+    false => undef,
+    default  => Class['puppetdb::server::validate_db'],
```
kbarber: Probably not worth doing it this time around, but worth thinking about later. If we ever get around to implementing a defined resource to wrap these ini settings or something, probably better to do it using that resource instead of inifile, since inifile could bleed into other modules and create more edges than necessary in the graph. Extraneous edges == bad sometimes, because it makes for a larger catalog.

kbarber:
@robinbowes this looks okay now, one thing we forgot to ask for before was the documentation changes for this new parameter in the README.md, can you add this? Sorry, shouldn't have noticed this earlier.

Also - you might want to squash your commits into 1 while you're at it.

robinbowes:
How's that?

kbarber:
@robinbowes perfect, thanks mate :-).

PR #159, opened 2015-01-02
(maint) don't use beaker for unit testing

wkalt:
Travis is failing on 1.8.7 due to some recent updates in rails. This should fix
the issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,11 +3,14 @@ source ENV['GEM_SOURCE'] || "https://rubygems.org"
 group :development, :test do
   gem 'rake'
   gem 'puppetlabs_spec_helper', :require => false
-  gem 'beaker',                 :require => false
-  gem 'beaker-rspec',           :require => false
-  gem 'serverspec',             :require => false
   gem 'rspec-puppet', '~> 1.0'
   gem 'puppet-lint',  '~> 1.1'
+end
+
+group :system_tests do
+  gem 'beaker',                :require => false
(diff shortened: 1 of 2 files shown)


kbarber:
So 'beaker', 'beaker-rspec', 'serverspec' aren't required for unit testing at all. Maybe instead of trying to do this, we just put those in their own Gemfile group, and exclude that in the .travis.yml? Take a look at how postgresql does it:

https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/.travis.yml
https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/Gemfile

wkalt:
@kbarber fixed

Gemfile
```diff
@@ -3,13 +3,21 @@ source ENV['GEM_SOURCE'] || "https://rubygems.org"
...
-  gem 'beaker',                 :require => false
-  gem 'beaker-rspec',           :require => false
-  gem 'serverspec',             :require => false
   gem 'rspec-puppet', '~> 1.0'
   gem 'puppet-lint',  '~> 1.1'
 end
 
+group :system_tests do
+  case RUBY_VERSION
+  when '1.8.7'
```
kbarber: I wouldn't even bother doing this, like I was saying last week, if beaker needs Ruby 1.9.3, then we should never run it on anything less. Be mindful that this versioning has nothing to do with our supported Ruby versions, since these are for acceptance tests, not for unit testing.

PR #160, opened 2015-01-05
add missing param manage_firewall

bastelfreak:
the param manage_firewall was added in an earlier release to puppetdb::server. The class isn't called directly but via the init.pp. the puppetdb class had already the param manage_firewall but this wasn't forwarded to puppetdb::server

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,6 +410,10 @@ Contents of your SSL certificate, as a string.
 
 Contents of your SSL CA certificate, as a string.
 
+####`manage_firewall`
+
+if true, puppet will manage your iptables rules for puppetdb via the [puppetlabs-firewall](https://forge.puppetlabs.com/puppetlabs/firewall) class.
+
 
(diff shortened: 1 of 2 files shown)


ekohl:
I can see the parameter exists here: https://github.com/puppetlabs/puppetlabs-puppetdb/blob/e56eb99146c69a5b07817b315ecb9813843569b6/manifests/init.pp#L56

ekohl:
Disregard that, I misread it.

electrical:
Test failure due to ruby dep issue: `Gem::InstallError: i18n requires Ruby version >= 1.9.3.` on the ruby 1.8.7 tests.

bastelfreak:
@electrical I guess this issue has nothing to do with the modified code and should be fixed in another PR?

electrical:
@bastelfreak yeah indeed. its unrelated to your change.

PR #162, opened 2015-01-14
add ability to manage postgres repo

bastelfreak:
the module triggers puppetlabs-postgresql which is able to handle yum/apt repos from the postgres devs. Usefull in cases that your operating system doesn't provide the newest release. this feature is disabled by default

just migrated my puppetdb postgres 9.2 from centos7 to 9.4, working fine with two puppet runs.


manifests/params.pp
```diff
@@ -16,13 +16,15 @@
...
-  $database_ssl           = false
-  $database_validate      = true
+  $database_host      = 'localhost'
+  $database_port      = '5432'
+  $database_name      = 'puppetdb'
+  $database_username  = 'puppetdb'
+  $database_password  = 'puppetdb'
+  $database_ssl       = false
+  $database_validate  = true
+  $postgres_version   = '9.4'
```
ajroetker: @kbarber should this be a 9.3 default for now?

manifests/database/postgresql.pp
```diff
@@ -1,14 +1,23 @@
...
+  $postgres_version     = $puppetdb::params::postgres_version,
 ) inherits puppetdb::params {
 
-  if $manage_server == true {
+  if $manage_server {
+    if $manage_package_repo {
+      class { '::postgresql::globals':
+        manage_package_repo => true,
+        version             => $postgres_version,
+        before              => Class['::postgresql::server'],
```
ajroetker: I think this ordering is implicit.

bastelfreak: as 9.4 is the newest version I thought it may be a good idea to use it. I will remove the before param

manifests/params.pp
kbarber: I don't mind the default being the latest, I can see this being another annoying place to constantly update however :-). Kind of wish there was a way to just say 'give me the latest always' instead. But hohum, yes 9.4 is fine.

kbarber:
@bastelfreak this will need updates to the README.md to cover the new parameters, can you please add these? We'll also need to functionally test this as well.

bastelfreak:
@kbarber updated the README.md. Can you offer any help or tutorials for the needed functionally test?

kbarber:
Don't worry @bastelfreak the functional test is on us to do today. We'll move this back into for review, thanks.

bastelfreak:
ah okay. let me know if you need additional changes or if I should do a rebase

PR #163, opened 2015-01-15
Allow puppetdb to be configure for masterless conf

Spredzy:
Currently the module only allows PuppetDB to be configured in a
agent/master setup configuration, by configuring the `[master]` section of
the puppet.conf. This commit allows one to use this module to configure
the `[main]` section of the puppet.conf so the module can configure nodes to
use PuppetDB in a masterless setup.

Doc about puppet.conf configure in masterless way available here
https://docs.puppetlabs.com/puppetdb/2.2/connect_puppet_apply.html#manage-puppetconf


manifests/master/config.pp
```diff
@@ -92,9 +94,10 @@
...
     class { 'puppetdb::master::report_processor':
-      puppet_conf => $puppet_conf,
-      enable      => $enable_reports,
-      require     => $strict_validation ? {
+      puppet_conf         => $puppet_conf,
+      puppet_conf_section => $puppet_conf_section,
+      enable              => $enable_reports,
+      require             => $strict_validation ? {
         true    => Puppetdb_conn_validator['puppetdb_conn'],
         default => Package[$terminus_package],
```
ajroetker: Would you still need the terminus in a masterless configuration?

ajroetker:
I'm tempted to say that maybe the storeconfigs and report_processor classes should be moved to a different namespace now, maybe under `puppetdb::puppet_conf::...`, and maybe changing the `master::config` class to be `puppetdb::consumer::config` (using a `$masterless = false` variable to determine the config section). @kbarber I'm not sure if we'd save all that for a major release? or if just copying the classes for now, instead of moving might be the thing to do to keep the master classes around.  My main concern is supporting masterless configuration in the master hierarchy :) love that it works with masterless though! :)

manifests/master/config.pp
Spredzy: @ajroetker ues you still need the terminus on every node https://docs.puppetlabs.com/puppetdb/2.2/connect_puppet_apply.html#step-2-install-terminus-plugins-on-every-puppet-node

Spredzy:
@ajroetker Thanks for reviewing. I agree about moving it out under another namespace, just wasn't make sure such a huge change will be accepted. If @kbarber agrees with the idea, I can make the needed changes.

Spredzy:
Up ?

kbarber:
@ajroetker we can't change the API on a non-major boundary. We'd need to create new classes or something and find some way to proxy from the old ones to the new ones to keep backwards compatibility. Having said that, we could almost start considering a 5.0 release at this point ... since we probably will need this for PDB 3.0.0 later on, however there is some complexity in just doing that right now - my concern is about the package naming really for the terminus, since that might introduce some lack of backwards compatibility with PDB 2.2.x. Maybe if this all gets too nasty we should cut a stable & master branch. This needs some more thought, and to be truly honest with you - introducing complexity like this right now is somewhat bad timing, due to the pressures we are under :-).

But I get what you are trying to say about naming, it does make some sense to create some differently named classes. Not sure if I like the term 'consumer' however, seems very generic, I'd just say 'puppet' or something like that, since thats what its trying to manage really.

kbarber:
@ajroetker so ... I'd shelve your idea in a ticket. I like it, its nice, but lets try and do it for a major release and when we have the time. Mark it as a 'fix for 5.0.0' or something like that, so we know exactly when it can be worked on. I think the naming problem here isn't terrible, and most people would be forgiving if they saw it. At least we know about it :-).

kbarber:
@ajroetker lets talk about this one today.

ajroetker:
@Spredzy if you wouldn't mind we thought that instead of having `$puppet_conf_section` maybe having a `$masterless = true` "switch" would be a better route for this because you could also use that variable to set other nice masterless defaults like `$manage_routes` to `false`. Looks good otherwise!

Spredzy:
@kbarber @ajroetker updated based on your comments. Tests fails only on ruby 1.8.7, I think it might be related to rspec-core 3.2.0 that came out 2 days ago based on the trace. I noticed the same issue on puppetlabs-mongodb module

kbarber:
@Spredzy yeah, I'm sending up a pinning patch now. Double check your patch with a grep, there are still some mentions of the old parameter btw.

Spredzy:
@kbarber Sorry I forgot to update the README. It has been fixed.

kbarber:
@Spredzy thanks mate, looking good.

Spredzy:
@kbarber @ajroetker Thanks to both of you for the reviews.

JohnTheodore:
Are there any simple instructions for standing up masterless? One of my issues is puppet masterless (with git pulls from a cronjob) is baked into our images. So when I'm running this masterless puppetdb stuff, it's inheriting from my puppet config /etc/puppet setup already on the box. So my puppet apply command has to be setup in a way to account for all of this...

Do you guys have some simple command/setup for spinning up puppetdb masterless for this?

PR #165, opened 2015-01-27
(PDB-1035) Set terminus package-name dynamically

ajroetker:
This PR sets the terminus package-name dynamically based on the version of PuppetDB we're using.


kbarber:
@ajroetker did you need some help with this one?

PR #166, opened 2015-02-04
Allow PostgreSQL db/user management to be disabled

rfletcher:
I'm configuring PuppetDB to connect to a third-party remote DB (Amazon RDS), so I've set `$manage_server` to `false`. That lead to this error from the Postgresql module (because the server isn't in my catalog):

> Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: Postgresql_psql[Check for existence of db 'puppetdb'] { require => Class[Postgresql::Server::Service] }, because Class[Postgresql::Server::Service] doesn't seem to be in the catalog

After digging in I found that despite disabling `$manage_server`, the module still tries to create the db/user.

I would probably have argued that `$manage_server` should affect db/user management too, but in the interest of backward compatibility I've instead added a new option to manage that independently (`$manage_db`), and preserved the current behavior by default.


kbarber:
This seems redundant to me, shouldn't you just not use the class instead? ie. just use puppetdb::server on its own? Thats pretty much how we planned on having this work.

I would recommend using the decomposed methodology in this case, instead of adding a brand new flag.

rfletcher:
Maybe! I'm not using the puppetdb module directly. Instead the puppet module we use is including it for me. I'm not sure if that module can be wrangled to including pupetdb::server by itself, but I'll look into it. Thanks for the advice.

PR #169, opened 2015-03-09
(bugfix) Use test_url in connection validator for puppetdb

ajroetker:
This commit addresses an issue where the provider for
puppetConnValidator was not using the test_url specified by the
resource. This commit passes the param from the initializer to the utils
class.


kbarber:
Testing for hipchat integration

PR #170, opened 2015-03-13
Add subparam passing for advanced database SSL options.

jonkelleyatrackspace:
In my organization there's a need to pass additional parameters when dealing with SSL.

This allows opting for custom SSL Factories, defining the SSL verify mode, and allows the power of advanced configurations when dealing with SSL.


manifests/server/read_database_ini.pp
```diff
@@ -62,7 +63,7 @@
       $subprotocol = 'postgresql'
 
       $subname = $database_ssl ? {
-        true    => "//${database_host}:${database_port}/${database_name}?ssl=true",
+        true    => "//${database_host}:${database_port}/${database_name}?ssl=true${database_subparam}",
```
ajroetker: In some of our other modules we use the following pattern for this
```
 $jdbc_ssl_properties   = "?ssl=true&sslfactory=org.postgresql.ssl.jdbc4.LibPQFactory&sslmode=verify-full&sslrootcert=${localcacert}"
```
That is to say maybe instead of the `database_ssl` bool there should just be a string args like `jdbc_ssl_properties` which we construct the subname from and `database_ssl` just equals `jdbc_ssl_properties == ''`.

ajroetker:
Closing in favor of ticketed work here https://tickets.puppetlabs.com/browse/PDB-1415.

PR #172, opened 2015-03-30
(PDB-1353) Use settings::confdir for puppet_confdir

johnduarte:
This commit adds conditionals to use `$settings::confdir` to introspect
the value for `$puppet_confdir` when available. This change is necessary
for compatibility with the path changes in Puppet 4.


kbarber:
This seems okay using puppetserver 1.0.8 (ie. non AIO pathing) on EL7.

PR #173, opened 2015-03-31
Sort routes hash in routes.yaml for Ruby 1.8.7

ghoneycutt:
Without this patch the ordering of the keys is not determiniate and will
continually change between puppet runs.


ghoneycutt:
This puts garbage in the routes.yaml and breaks your puppet master

PR #175, opened 2015-04-07
(MODULES-1917) Ignore `._` files when building the module tarball

ajroetker:
Version bump to build a version without `._` files.


rbrw:
@kbarber @ajroetker Seems fine for now, and I imagine we may want to come up with some broader way of making sure we always specify COPYFILE_DISABLE on OS X.

PR #176, opened 2015-04-09
Enable the module to manage entries in $confdir/config.ini

buzzdeee:
with regard to the command-processing section.

Added new class server/config_ini.pp to manage contents of the config.ini.
Three new parameters added:
  * command_threads
  * store_usage
  * temp_usage

All three default to 'undef'. This makes sure (potential) custom settings
done to that file with regard to above three variables are 'absent',
and let PuppetDB built-in defaults take care.

Documentation to the README.md added, as well as unit tests.

My use-case was, that I have on some nodes a too small /var partition,
so I had to lower the values of store-usage and temp-usage in the config.ini
manually.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -55,6 +55,9 @@ class puppetdb::server (
   $manage_firewall          = $puppetdb::params::manage_firewall,
   $java_args                = $puppetdb::params::java_args,
   $max_threads              = $puppetdb::params::max_threads,
+  $command_threads          = $puppetdb::params::command_threads,
+  $store_usage              = $puppetdb::params::store_usage,
+  $temp_usage               = $puppetdb::params::temp_usage,
 ) inherits puppetdb::params {
 
   # Apply necessary suffix if zero is specified.
@@ -114,6 +117,13 @@ class puppetdb::server (
       ssl_port       => $ssl_listen_port,
       open_ssl_port  => $open_ssl_listen_port,
     }
+  }
+
+  class { 'puppetdb::server::config_ini':
+    command_threads => $command_threads,
+    store_usage     => $store_usage,
+    temp_usage      => $temp_usage,
+    confdir         => $confdir,
   }
 
   class { 'puppetdb::server::database_ini':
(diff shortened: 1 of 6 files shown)


wkalt:
this looks good to me

manifests/server.pp
```diff
@@ -55,6 +55,9 @@
   $manage_firewall          = $puppetdb::params::manage_firewall,
   $java_args                = $puppetdb::params::java_args,
   $max_threads              = $puppetdb::params::max_threads,
+  $command_threads          = $puppetdb::params::command_threads,
```
ajroetker: Since these all default to undef, you could probably just explicitly set them to undef here.

manifests/server/config_ini.pp
```diff
@@ -0,0 +1,51 @@
...
+  $temp_usage      = $puppetdb::params::temp_usage,
+  $confdir         = $puppetdb::params::confdir,
+) inherits puppetdb::params {
+
+  # Set the defaults
+  Ini_setting {
+    path    => "${confdir}/config.ini",
+    ensure  => present,
+    section => 'command-processing',
+  }
```
ajroetker: You may need to manage the file resource itself as well, since a PDB package might not have it there by default.

ajroetker:
:+1: aside from the `undef` comment which I could live without!

buzzdeee:
I added a file resource to the server/config_ini.pp file to take care that it exists before trying to manipulate it. On OpenBSD, its owned by root:wheel, however, that is usually the user puppet runs on.
I was unsure, if I should explicitly manage the owner/group of the file. i.e. owner => root, group => 0.
since the permissions allow everyone to read, i.e. 0644, so puppetdb should be fine with it.


I kept the parameter definitions to undef in param.pp, because they are passed in in init.pp, and server.pp. If everyone ever wants to change them, it can be done centrally in init.pp, so I hope that is fine.

Unfortunately, the travis tests currently fail, because of updates/changes in the "concat" modules git repo. I didn't wanted to mess with the specs, therefore I haven't done anything in that regard ;)

kbarber:
Looks like we pull in another dependency for concat now, it wraps a native tool from electrical now: https://github.com/puppetlabs/puppetlabs-postgresql/pull/612/files. I guess we can either pin, or bump.

kbarber:
Trying to fix the testing issues here: https://github.com/puppetlabs/puppetlabs-puppetdb/pull/177

buzzdeee:
rebased, and specs seem to be all green again.

PR #178, opened 2015-04-29
Add a parameter to manage or not the PostgreSQL database

roidelapluie:


roidelapluie:
This PR is obsolete, if you want not to manage the db then use include puppetdb::server instead of include puppetdb

roidelapluie:
But then params would not be initalized. repoening.

roidelapluie:
server.pp inherits params as well so this pr is indeed not wanted.

PR #179, opened 2015-04-30
read_database.ini ordered before the puppetdb package

fiddyspence:
On EL6, Puppet 3.7.4 I get this error:

Error: Could not set 'file' on ensure: No such file or directory - /etc/puppetdb/conf.d/read_database.ini at 43:/platform/puppet/environment/production/modules/puppetdb/manifests/server/read_database_ini.pp

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -144,6 +144,7 @@ class puppetdb::server (
     conn_keep_alive     => $conn_keep_alive,
     conn_lifetime       => $conn_lifetime,
     confdir             => $confdir,
+    require             => Package[$puppetdb_package],
     notify              => Service[$puppetdb_service],
   }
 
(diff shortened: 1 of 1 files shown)


manifests/server.pp
```diff
@@ -161,6 +161,7 @@
     conn_keep_alive     => $read_conn_keep_alive,
     conn_lifetime       => $read_conn_lifetime,
     confdir             => $confdir,
+    require             => Package[$puppetdb_package],
```
ajroetker: If this require is needed here, will it be needed in the other config file resources as well?

kbarber: I'd say so, I think we should produce a graph and manually double check this maybe. I can see a few just in code, wouldn't hurt to visually check it.

kbarber:
I'm going to close this in favour of #180, this follows the same style for keeping order as before.

PR #181, opened 2015-05-15
(PDB-1035) Add default PuppetDB root context

ajroetker:
This commit changes the terminus configuration to use the new root
prefix for PuppetDB and changes the connection validator to do the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,15 @@ You can also manually trigger puppet runs on the nodes in the correct order (Pos
 Upgrading
 ---------
 
+###Upgrading from 4.x to version 5.x
+
+Significant parameter changes are listed below:
+
+* The PuppetDB module now supports PuppetDB 3.0.0 by default
+* If you want to use 5.x of the module with PuppetDB 2.x, you'll need to set the `test_url => /v3/version` and either `puppetdb_version => 2.y.z` or `terminus_package => 2.y.z`
+
+See the CHANGELOG file for more detailed information on changes for each release.
+
 ###Upgrading from 3.x to version 4.x
 
 For this release, all dependency versions have been bumped to their latest. Significant parameter changes are listed below:
@@ -557,7 +566,7 @@ Conditionally manages the PostgresQL server via `postgresql::server`. Defaults t
 
 ####`test_url`
 
-The URL to use for testing if the PuppetDB instance is running. Defaults to `/v3/version`.
+The URL to use for testing if the PuppetDB instance is running. Defaults to `/pdb/meta/v1/version`.
 
 ####`manage_package_repo`
 
--- a/lib/puppet/util/puppetdb_validator.rb
+++ b/lib/puppet/util/puppetdb_validator.rb
@@ -10,7 +10,7 @@ module Puppet
       attr_reader :test_path
       attr_reader :test_headers
 
-      def initialize(puppetdb_server, puppetdb_port, use_ssl=true, test_path = "/v3/version")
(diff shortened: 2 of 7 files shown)


lib/puppet/provider/puppetdb_conn_validator/#puppet_https.rb#
```diff
@@ -0,0 +1,72 @@
+# See: #10295 for more details.
```
kbarber: Invalid file committed.

lib/puppet/util/puppetdb_validator.rb
```diff
@@ -10,7 +10,7 @@ class PuppetdbValidator
       attr_reader :test_path
       attr_reader :test_headers
 
-      def initialize(puppetdb_server, puppetdb_port, use_ssl=true, test_path = "/v3/version")
+      def initialize(puppetdb_server, puppetdb_port, use_ssl=true, test_path = "/pdb/query/v4/version")
```
kbarber: Okay, this makes the module defaulting to behaviour not yet released, so it will break stuff. We've threatened to do this for a while, but I think we need to coordinate this properly, we might need to cut a new set of branches to handle the new state as we start to make breaking changes (ie. stable/master etc.). Lets have a chat about this before merging, make sure we have a plan for module v5.0.

spec/unit/classes/master/config_spec.rb
```diff
@@ -26,53 +26,53 @@
...
-        :puppetdb_port => '1234'
-      } end
+      let(:params) do {:puppetdb_port => '1234'} end
 
       it { should contain_puppetdb_conn_validator('puppetdb_conn').with(
-        :puppetdb_port => '1234',
-        :use_ssl => 'false')
-      }
+                    :puppetdb_port => '1234',
+                    :use_ssl => 'false')}
```
rbrw: Not crititcal, but I might be tempted to leave out the whitespace changes or put them in a separate patch so that this one is focused on the /pdb/query changes.

kbarber: This is more like a clojure formatting anyway, not ruby. I'd stay with what we had TBH, its more idiomatic.

kbarber: Meh, we're guilty of clojure like formatting throughout all this code anyway. Lol.

kbarber:
@ajroetker I think this patch needs docs updates calling out how to use the old terminus name, and the old test url if one is still on the old version of PDB. Since users who have just used ensure => installed, won't be able to get the benefits of your auto-detection, and may still be using PDB 2.3.x.

kbarber:
This duplicates the work here: https://tickets.puppetlabs.com/browse/PDB-1035

kbarber:
@ajroetker this still needs doc updates.

ajroetker:
@kbarber I know I was waiting for the the version/cmd patch because this will have to change again after that goes in

kbarber:
@ajroetker okay, makes sense.

README.md
```diff
@@ -139,6 +139,15 @@ You can also manually trigger puppet runs on the nodes in the correct order (Pos
...
 
+###Upgrading from 4.x to version 5.x
+
+Significant parameter changes are listed below:
+
+* The PuppetDB module now supports PuppetDB 3.0.0 by default
+* If you want to use 5.x of the module with PuppetDB 2.x, you'll need to set the `test_url => /v3/version` and either `puppetdb_version => 2.y.z` or `terminus_package => 2.y.z`
+
+See the CHANGELOG file for more detailed information on changes for each release.
+
```
wkalt: v3/version is still mentioned a number of times in this file

PR #182, opened 2015-05-28
(maint) Allow database_validation to be false

ajroetker:
This commit fixes a bug where if a user tried to turn off database
validation there would be a `before` statement requiring an absent
validation resource.


kbarber:
@ajroetker probably target this at stable, in case we need to release it.

ajroetker:
will do

PR #185, opened 2015-06-10
(PDB-1455) Provide mechanism for modifying default HSQLDB path

kbarber:
This provides an override mechanism for providing a custom path to HSQLDB.

This allows users to provide any path to their HSQLDB database. It is needed
specifically for managing the as-yet-unreleased version of PuppetDB as well.


manifests/params.pp
```diff
@@ -70,7 +70,7 @@
   case $::osfamily {
     'RedHat', 'Suse', 'Archlinux': {
       $confdir              = '/etc/puppetdb/conf.d'
-      $embedded_subname     = 'file:/var/lib/puppetdb/db/db;hsqldb.tx=mvcc;sql.syntax_pgs=true'
+      $database_path        = '/var/lib/puppetdb/db/db'
```
kbarber: The embedded_subname was never a configurable parameter, so I've switched the focus here over to just the path - now we fill in the remaining parts of subname inside database_ini.pp in only one place (since it never should be different per OS anyway).

spec/unit/classes/server/database_ini_spec.rb
```diff
@@ -0,0 +1,140 @@
+require 'spec_helper'
```
kbarber: There was no coverage here, so I've added default coverage+tests for this particular change.

README.md
```diff
@@ -257,6 +257,10 @@ Setting up proper trust- and keystores has to be managed outside of the puppetdb
 
 If true, the module will attempt to connect to the database using the specified settings and fail if it is not able to do so. (defaults to true)
 
+####`database_path`
```
ajroetker: I'd prefer this be called something more like `embedded_database_path` just to be explicit about the case it's used for, it will also signal to users that they need not care about it.

ajroetker:
:+1: modulo the name change, let me know what you think about that

README.md
kbarber: How about database_embedded_path? Thats how all the other variables are named.

spec/unit/classes/server/database_ini_spec.rb
ajroetker: Nice!

README.md
kbarber: Done, you'll want to git diff -w, since we still use the aligning arrows/equals style here and I had to line up a lot of stuff since database_embedded_path is long :-).

ajroetker:
lgtm tests passed for me locally, I'll merge when they go green

kbarber:
@ajroetker green now.

PR #189, opened 2015-06-17
Be compatible with stdlib v5

ghoneycutt:
According to stdlib's documentation, v5 should be compatible here.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -86,7 +86,7 @@
         },
         {
             "name": "puppetlabs/stdlib",
-            "version_requirement": ">= 4.2.2 <5.0.0"
+            "version_requirement": ">= 4.2.2 <6.0.0"
         }
     ]
 }


kbarber:
@ghoneycutt I'm going to take a wild guess, correct me if I'm wrong here: Is the reasoning behind support this future version something to do with the support matrix and the lack of documented 4 support? I had a chat to @DavidS and he raised this to correct the docs, afaik 4.x does support puppet 4, there has been a PR raised to correct the docs and metadata: https://github.com/puppetlabs/puppetlabs-stdlib/pull/475

Just a little surprising to see us declaring support for something that isn't released :-).

ghoneycutt:
Maybe a little too forward thinking on my part :)

Thought I'd address it while I was mucking around with the other metadata bits and Puppetv4 support in my other PR's. According to the docs, it should be compatible with v5, though since it has not been released, you can close this.

kbarber:
Okay, lets close for now.

PR #193, opened 2015-06-18
(PDB-1523) Add pathing for AIO, defaulting to non-AIO for older PDBs

ajroetker:
This commit adds a globals class to PuppetDB which allows us to change
the param defaults for the module depending on what version of PuppetDB
they are using (similar to the PostgreSQL module).
This commit also changes the default PuppetDB 3.x configuration pathing
to assume AIO Puppet.


ajroetker:
@kbarber here's a merge-able prototype when you have a chance to take a look

kbarber:
PDB-1523 is the ticket mate, btw.

README.md
```diff
@@ -144,23 +144,24 @@ Upgrading
 Significant parameter changes are listed below:
 
 * The PuppetDB module now supports PuppetDB 3.0.0 by default
-* If you want to use 5.x of the module with PuppetDB 2.x, you'll need to set `puppetdb_version => 2.y.z` or `terminus_package => 2.y.z`
-* The `puppetdb::master:puppetdb_conf` class has added a `$legacy_terminus` parameter which will be set the correct default if you set `puppetdb_version => 2.y.z` or `terminus_package => 2.y.z`.
-* The default `test_url` for the `PuppetDBConnValidator` has also been chaged to `/pdb/meta/v1/version` but will also be set the correct default if you set `puppetdb_version => 2.y.z` or `terminus_package => 2.y.z`.
+* If you want to use 5.x of the module with PuppetDB 2.x, you'll need to use the new `puppetdb::globals` class to set the version of PuppetDB you're using explicitly.
```
kbarber: We probably want a prominent primary section in the docs for the globals class now perhaps.

kbarber:
Rerunning travis, now we have Puppet 4 & future parser testing.

kbarber:
Actually, rerunning won't be enough it seems, it still tests just the old stuff instead of the merge for whatever reason. Might want to do a rebase for good luck :-).

README.md
ajroetker: Ah yes that would definitely be a good idea

README.md
```diff
@@ -204,18 +207,31 @@ Usage
...
 
     class { 'puppetdb': }
 
-**Parameters within `puppetdb`:**
+**parameters within `puppetdb`:**
 
 ####`listen_address`
 
-The address that the web server should bind to for HTTP requests (defaults to `localhost`.'0.0.0.0' = all).
+the address that the web server should bind to for http requests (defaults to `localhost`.'0.0.0.0' = all).
```
wkalt: looks like this doc got lowercased accidentally

kbarber:
@hunner is the globals pattern still a Good Idea (tm) these days?

README.md
```diff
@@ -204,18 +207,31 @@ Usage
...
+
 ### puppetdb
-The `puppetdb` class is intended as a high-level abstraction (sort of an 'all-in-one' class) to help simplify the process of getting your puppetdb server up and running. It wraps the slightly-lower-level classes `puppetdb::server` and `puppetdb::database::*`, and it'll get you up and running with everything you need (including database setup and management) on the server side.  For maximum configurability, you may choose not to use this class.  You may prefer to use the `puppetdb::server` class directly, or manage your puppetdb setup on your own.
+the `puppetdb` class is intended as a high-level abstraction (sort of an 'all-in-one' class) to help simplify the process of getting your puppetdb server up and running. it wraps the slightly-lower-level classes `puppetdb::server` and `puppetdb::database::*`, and it'll get you up and running with everything you need (including database setup and management) on the server side.  for maximum configurability, you may choose not to use this class.  you may prefer to use the `puppetdb::server` class directly, or manage your puppetdb setup on your own.
```
kbarber: Same here lower case 't'.

manifests/init.pp
```diff
@@ -40,7 +40,6 @@
   $conn_keep_alive          = $puppetdb::params::conn_keep_alive,
   $conn_lifetime            = $puppetdb::params::conn_lifetime,
   $puppetdb_package         = $puppetdb::params::puppetdb_package,
-  $puppetdb_version         = $puppetdb::params::puppetdb_version,
```
kbarber: So you've removed the ability to tweak this on a per class basis from puppetdb & puppetdb::server, but not from puppetdb::master::config, what direction were you thinking globally?

ajroetker: Changed it so it's not configurable through `master` either.

hunner:
@kbarber Yeah, the globals pattern works as far as anything would when a user needs to affect params in ways that are not otherwise achievable. It'd be nice if the parser could give help on which classes must be parsed ahead of time, but alas we can't.

README.md
```diff
@@ -204,14 +207,27 @@ Usage
 
 PuppetDB supports a large number of configuration options for both configuring the puppetdb service and connecting that service to the puppet master.
 
+### puppetdb::globals
+the `puppetdb::globals` class is intended to provide similar functionality to the `puppetlabs-postgresql`'s `postgresql::globals` class providing a top-level entry-point into the module so that we can properly set defaults for the `puppetdb::params` class based on the version of `puppetdb` you are using. This setting defaults to `present`.
```
kbarber: T

README.md
```diff
@@ -204,14 +207,27 @@ Usage
 
 PuppetDB supports a large number of configuration options for both configuring the puppetdb service and connecting that service to the puppet master.
 
+### puppetdb::globals
+the `puppetdb::globals` class is intended to provide similar functionality to the `puppetlabs-postgresql`'s `postgresql::globals` class providing a top-level entry-point into the module so that we can properly set defaults for the `puppetdb::params` class based on the version of `puppetdb` you are using. This setting defaults to `present`. 
+
+you must declare the class to use it:
```
kbarber: Y

README.md
```diff
@@ -204,14 +207,27 @@ Usage
...
+
+    class { 'puppetdb::globals': }
+
+**parameters within `puppetdb::globals`:**
```
kbarber: P

README.md
```diff
@@ -204,14 +207,27 @@ Usage
...
+
+####`version`
+
+The version of the `puppetdb` package that should be installed.  You may specify an explicit version number, 'present', or 'latest' (defaults to 'present').
+
 ### puppetdb
 The `puppetdb` class is intended as a high-level abstraction (sort of an 'all-in-one' class) to help simplify the process of getting your puppetdb server up and running. It wraps the slightly-lower-level classes `puppetdb::server` and `puppetdb::database::*`, and it'll get you up and running with everything you need (including database setup and management) on the server side.  For maximum configurability, you may choose not to use this class.  You may prefer to use the `puppetdb::server` class directly, or manage your puppetdb setup on your own.
 
-You must declare the class to use it:
+you must declare the class to use it:
```
kbarber: Y

README.md
```diff
@@ -204,14 +207,27 @@ Usage
...
 
     class { 'puppetdb': }
 
-**Parameters within `puppetdb`:**
+**parameters within `puppetdb`:**
```
kbarber: P

kbarber:
Functionally testing this now - first pass is making sure it can still install legacy, second pass is making sure it can install new pdb.

README.md
```diff
@@ -144,23 +144,26 @@ Upgrading
...
-* The `puppetdb::master:puppetdb_conf` class has added a `$legacy_terminus` parameter which will be set the correct default if you set `puppetdb_version => 2.y.z` or `terminus_package => 2.y.z`.
-* The default `test_url` for the `PuppetDBConnValidator` has also been chaged to `/pdb/meta/v1/version` but will also be set the correct default if you set `puppetdb_version => 2.y.z` or `terminus_package => 2.y.z`.
+* If you want to use 5.x of the module with PuppetDB 2.x, you'll need to use the new `puppetdb::globals` class to set the version of PuppetDB you're using explicitly. The ability to configure the version has been therefore moved out of the `puppetdb` and `puppetdb::server` classes.
 For example if your config looked like this before:
 ~~~ruby
-class { 'puppetdb::master::config':
-      puppetdb_server => 'foo.example.com',
-      puppetdb_version => present,
+class {'puppetdb':
+  puppetdb_version => '2.3.5',
```
kbarber: This isn't correct for say, el7, right now you need the version string to be more like `2.3.5-1.el7`. I think debian needs the -debian1 suffix or something also.

kbarber:
Legacy case seems fine, except for that 1 point about the versioning label. Testing the 3.0 case now.

README.md
```diff
@@ -144,23 +144,26 @@ Upgrading
...
 ~~~
 and you'd still like to use the module with PuppetDB 2.3.5, all you'd have to change would be:
 ~~~ruby
-class { 'puppetdb::master::config':
-      puppetdb_server => 'foo.example.com',
-      terminus_package => '2.3.5',
+class { 'puppetdb::globals':
+  version => '2.3.5',
 }
+include puppetdb
```
kbarber: Purely a fine point, and a matter of a style, but we should promote the class { } style instead of include generally, I prefer it for instruction, so its easier to explain to users how to add parameters versus the include style which requires a bigger change.

README.md
ajroetker: Cool I'll tidy up the examples.

manifests/params.pp
```diff
@@ -11,9 +11,12 @@
   $disable_ssl               = false
   $open_ssl_listen_port      = undef
   $postgres_listen_addresses = 'localhost'
-
-  $database                  = 'postgres'
+  
+  $puppetdb_version          = $puppetdb::globals::version
+  $database                  = $puppetdb::globals::database
   $manage_dbserver           = true
+  $manage_pg_repo            = false
```
kbarber: So ... I wonder if its time to switch this to true. Right now if you install the module on el7 it fails, because it tries to install 9.2, this flag is actually ignored in a manage_repos = false case btw (different issue, but important to note also).

I feel like this is a major barrier for new users, so maybe worth it?

ajroetker: I agree, if we want to be in a situation where the module "just works" for most of our users, then more often than not, we'll want to manage the pg_repos by default since most systems don't have 9.3/9.4 installed.

mullr: Seems like most puppet modules *don't* manage repos by default, do they?

kbarber: Shall we do this as part of this PR? Or you wanna create a separate one? Also there is that bug with the manage_repos logic not taking a version if it isn't specified, I can't recall if that was done for a reason.

ajroetker: I'd prefer to do it in another PR, I'll make a ticket for it. To Russ's comment, what we're doing with the class in question seems to be trying to give users the least barrier to entry in terms of getting a PuppetDB up and running, most intermediate/advanced users won't even be interacting with the class at all and will probably be managing postgres via it's own module. I think asking novice users to go look through documentation to find that they need to turn manage_repos on rather than telling advanced users they need to turn it off is probably not where we want to be.

ajroetker: ping @mullr

mullr: It don't disagree, but as a wise man once said: "standard is better than better"

kbarber: @mullr I think in this case though, it won't work without tweaking for the common RHEL users, that a large part of the user base. That means for PuppetDB, they'll need their own repo managed anyway. Right now the postgresql module can manage these repos to supply a 9.4, I think it's not unrealistic to think that most users will want this. Otherwise, they get a failure, and then have to go off to the docs to turn this on, I can't think of another way they'd want to solve this and have a working PDB stack so it seems like a reasonable default approach. My 2p anyway.

mullr: @kbarber @ajroetker ok, will definitely defer to you guys on this

kbarber: FWIW, this will affect Debian 6 also, but Debian 7 looks like it has a 9.4.

Something more to add, we want to promote the use of PGDG generally so we can promote latest postgres usage, today its 9.4, tomorrow its 9.5/9.6. When 9.6 for example is the one we want users to use, there is something to be said about having our users go in the direction of PGDG, and just be able to use it, without having to go back again and think about manage_repos etc.

kbarber:
Okay, despite the fact that I need this to make it work:

{code}
class { 'puppetdb':
  manage_package_repo => true,
  postgres_version => '9.4',
}
class { 'puppetdb::master::config': }
{code}

This PR works with 3.0.0. @ajroetker are you creating a ticket to cover this manage_package_repo work? We should interrupt the sprint with it also, that is, we probably want to ship 5.0 with that default change.

README.md
```diff
@@ -204,10 +207,23 @@ Usage
 
 PuppetDB supports a large number of configuration options for both configuring the puppetdb service and connecting that service to the puppet master.
 
+### puppetdb::globals
+The `puppetdb::globals` class is intended to provide similar functionality to the `puppetlabs-postgresql`'s `postgresql::globals` class providing a top-level entry-point into the module so that we can properly set defaults for the `puppetdb::params` class based on the version of `puppetdb` you are using. This setting defaults to `present`.
```
wkalt: "intended to provide similar functionality to the `puppetlabs-postgresql`'s `postgresql::globals` class providing "

seems a bit misworded. Maybe change to 

"intended to provide similar functionality to the `postgresql::globals` class in the `puppetlabs-postgresql` module by exposing "

or something

ajroetker:
@kbarber yep I'll make a ticket asap and pull it in and start working on it

ajroetker:
@kbarber what failure are you seeing with the postgres_version not defaulting properly? I don't see anything in the code that would suggest a problem

kbarber:
@ajroetker if manage_repos is off, it won't pass through the postgres version: https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/database/postgresql.pp#L13-L25.

Its not specific to this PR, its more related to the other PR.

kbarber:
Merged manually.

PR #197, opened 2015-06-24
(PDB-1657) Manage Postgres repos by default

ajroetker:
This commit moves to managing the Postgres repos by default. The
reason for this is so that the `puppetdb` class will "just work" by
default on most systems, because PostgreSQL 9.4 is not installed on 6/7
of the distros we support we need to manage the repos so we can install
the latest version.


README.md
```diff
@@ -144,6 +144,7 @@ Upgrading
 Significant parameter changes are listed below:
 
 * The PuppetDB module now supports PuppetDB 3.0.0 by default
+* The PuppetDB module now manages Postgres repos by default, to turn this behavior off, set `manage_package_repo` to false.
```
wkalt: first comma should be a period with a new sentence

README.md
```diff
@@ -598,11 +599,11 @@ The URL to use for testing if the PuppetDB instance is running. Defaults to `/pd
 
 ####`manage_package_repo`
 
-if this is true, the official postgres.org repo will be added and postgres won't be installed from the regular repository.
+If this is true, the official postgres.org repo will be added and postgres won't be installed from the regular repository. This setting defaults to `true`.
 
 ####`postgres_version`
 
-if the postgres.org repo is installed, you can install several versions of postgres, this currently defaults to 9.4 which is the latest stable version.
+If the postgres.org repo is installed, you can install several versions of postgres, this currently defaults to 9.4 which is the latest stable version.
```
wkalt: last comma -> period

wkalt:
a couple tiny comments above -- looks good to me otherwise

PR #199, opened 2015-06-29
added local puppetdb_version variable to puppetdb::server class

SimonHoenscheid:
to make it configurable if multinode setup is needed, fallback to params.pp version

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -39,6 +39,7 @@ class puppetdb::server (
   $puppetdb_service_status  = $puppetdb::params::puppetdb_service_status,
   $puppetdb_user            = $puppetdb::params::puppetdb_user,
   $puppetdb_group           = $puppetdb::params::puppetdb_group,
+  $puppetdb_version         = $puppetdb::params::puppetdb_version,
   $read_database            = $puppetdb::params::read_database,
   $read_database_host       = $puppetdb::params::read_database_host,
   $read_database_port       = $puppetdb::params::read_database_port,
@@ -108,7 +109,7 @@ class puppetdb::server (
   }
 
   package { $puppetdb_package:
-    ensure => $puppetdb::params::puppetdb_version,
+    ensure => $puppetdb_version,
     notify => Service[$puppetdb_service],
   }
 


SimonHoenscheid:
we use a multinode setup. if the puppetdb::server class is assigned to one node, you can't assign the puppetdb:version.

```
class { 'puppetdb::server':
  puppetdb_version => '2.3.5...'
}
```

ajroetker:
@SimonHoenscheid Is there something preventing the use of the `puppetdb::globals` class here? You should be able to use `class { 'puppetdb::globals' : version => '2.3.5...' }` here, unless you're wanting to have multiple version of PuppetDB

SimonHoenscheid:
If I follow the documentation for a multinode System, I should do it, like I described it above.

ajroetker:
@SimonHoenscheid I'll have another look at the README and other documentation to make sure it's more clear, but basically with the release of Puppet 4 we had to change a lot of the defaults throughout the module (mostly pathing related), so we changed the way you specify the PuppetDB version so that we can override default params in the params class (i.e. from the globals class). The master branch's README and CHANGELOG should reflect these changes but we may have missed some docs. If you use this patch the way you have it now, I think you'll end up with some unintended pathing issues.

SimonHoenscheid:
I did a configuration with this patch today. had no problems. if you do not declare the variable the value from params is used, which points to globals.pp. Many parameters in puppetdb::server work the same way.

ajroetker:
@SimonHoenscheid are you using Puppet 3 or Puppet 4? With Puppet 4 this patch should work fine, but the method you're using for passing the version will break the module's Puppet 3 compatibility.

SimonHoenscheid:
I used puppet 3.7.5

ajroetker:
@SimonHoenscheid So here in the docs we have examples of the globals class https://github.com/puppetlabs/puppetlabs-puppetdb/blob/8eb9f676694196b654c30d98b1ba5306a9c0f5c4/README.md#upgrading. I'm not sure how this patch works if you pass a different version in the `puppetdb::server` class from the one that gets passed in `puppetdb::globals`. The globals class defaults to using present, and the params class will use that value and default to using Puppet 4 pathing for the config files and the puppet.conf file. The `puppetdb::server` class will then subsequently use the wrong path for the `database_ini` etc unless you pass those parameters explicitly.

SimonHoenscheid:
@ajroetker now all the parts get together. I would suggest to modify that part of the documentation, it seems outdated:

https://github.com/puppetlabs/puppetlabs-puppetdb/blob/8eb9f676694196b654c30d98b1ba5306a9c0f5c4/README.md#multiple-node-setup

kbarber:
@SimonHoenscheid what is precisely outdated in that example? The example does work as it says doesn't it? I presume you mean you want an example that shows how to modify the globals version - which to me is a different example from the original 'basic' one. We do have a section for the globals class already, I guess it could be expanded with a small example as well there perhaps.

kbarber:
@SimonHoenscheid ping here also :-).

kbarber:
Closing this for now, please address the comments above and re-open when you are ready.

PR #200, opened 2015-06-30
fixed serval puppet-lint warnings

SimonHoenscheid:
Fixed serval puppet-lint warnings. selector inside resource block still remaining


manifests/database/postgresql.pp
```diff
@@ -8,8 +8,8 @@
   $manage_server        = $puppetdb::params::manage_dbserver,
   $manage_package_repo  = $puppetdb::params::manage_pg_repo,
   $postgres_version     = $puppetdb::params::postgres_version,
-) inherits puppetdb::params {
-
+) {
+  include puppetdb::params
```
kbarber: Hmm, what is the value gained by changing this pattern to one using include? The style guide says this particular case is okay fwiw: https://docs.puppetlabs.com/guides/style_guide.html#class-inheritance. I think also this might be the reason some of the unit tests are failing.

manifests/master/storeconfigs.pp
```diff
@@ -1,9 +1,10 @@
-# This class configures the puppet master to enable storeconfigs and to use
+# inherits puppetdb::paramsThis class configures the puppet master to enable storeconfigs and to use
```
kbarber: Typo

kbarber:
@SimonHoenscheid the unit tests are failing on this patch across the board, see the 'details' button above next to "travis ci build failed" ^^, at first glance it seems like the change to the params pattern has made the variables in the class declaration from params out of scope (or something like this).

kbarber:
@SimonHoenscheid ping :-).

SimonHoenscheid:
I'm not sure, if I will find time to fix that soon. I will close that PR and split it up into little parts.

PR #203, opened 2015-08-05
PuppetDB version can be provided from main puppetdb class instead fro…

rameshmimit:
…m globles, as version comparison condition is getting failed for redhat/centos system which leads to wrong configuration location

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -1,5 +1,4 @@
 class puppetdb::globals (
-  $version                      = 'present',
   $database                     = 'postgres',
   ) {
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -40,6 +40,7 @@ class puppetdb (
   $conn_keep_alive          = $puppetdb::params::conn_keep_alive,
   $conn_lifetime            = $puppetdb::params::conn_lifetime,
   $puppetdb_package         = $puppetdb::params::puppetdb_package,
+  $puppetdb_version         = $puppetdb::params::puppetdb_version,
   $puppetdb_service         = $puppetdb::params::puppetdb_service,
   $puppetdb_service_status  = $puppetdb::params::puppetdb_service_status,
   $puppetdb_user            = $puppetdb::params::puppetdb_user,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -12,7 +12,7 @@ class puppetdb::params inherits puppetdb::globals {
   $open_ssl_listen_port      = undef
   $postgres_listen_addresses = 'localhost'
 
-  $puppetdb_version          = $puppetdb::globals::version
+  $puppetdb_version          = '2.3.6-1.el6'
   $database                  = $puppetdb::globals::database
   $manage_dbserver           = true
   $manage_pg_repo            = true


rameshmimit:
Adding provision to provide version no. for puppetdb

kbarber:
I'm not sure we can take this patch. The globals pattern & requesting version globally is there for a reason, to ensure we can set the version globally for all the various classes & resources available. Removing that for version breaks this goal, not only that it breaks backwards compatibility.

Also, I don't see us ever wanting to pin against the older 2.3.x version of PuppetDB. The module by default should be installing the whatever is the latest by default from scratch (3.0.2 at this point, but that moves), and keeping what is installed (if something already exists). It should however support 2.3.6, if desired. We certainly are able to do this during our acceptance tests.

If there is some sort of bug with the version comparison, I'm sure it can be solved without breaking the API or making a major change in behaviour, perhaps we should work towards that goal instead @rameshmimit. If you can raise another patch that requires less changes to the patterns in place or perhaps (and just fixes the bug) or raise a bug that describes the issue, that would be helpful.

earsdown:
Ken - just got bitten by this bug too - see https://tickets.puppetlabs.com/browse/MODULES-2334. EL7 variants are extremely popular operating systems, and at time of writing, this module requires production instances to pin against a specific 2.3.x versions of puppetdb. The best fix would be to get this feature working: https://tickets.puppetlabs.com/browse/PUP-1519, but until it is, I don't think this design pattern is viable.

kbarber:
FWIW, we test on EL7 and most popular variants daily using the module. @earsdown the problem sounds like you aren't using the PC1 repos: https://docs.puppetlabs.com/guides/puppetlabs_package_repositories.html#using-puppet-collections

Thats where version 3 & the puppetdb-termini package lives. We don't manage the puppetlabs repo from this module, so users are left doing this themselves. This could either be managed (I'm very wary, since those repos might already being managed by another module for example - pl repos are "bigger" than this module) or we just make sure this is more explicit in the documentation.

earsdown:
I think users will just follow the puppet installation documentation: https://docs.puppetlabs.com/guides/install_puppet/install_el.html ... which is somewhat silent on the use of puppet collection repos.... besides, I'm not sure how excited I am to move on to puppet 4 for production systems...

earsdown:
Thanks for the quick reply btw!

kbarber:
@earsdown yeah its bloody squirrely, but note the quote in that doc:

"Note: This document covers open source releases of Puppet version 3.8 and lower. You might also want instructions for installing Puppet Enterprise or installing Puppet 4.0 or newer."

And then links to specific docs for Puppet 4: https://docs.puppetlabs.com/puppet/latest/reference/install_pre.html

We're in that in-between world of old v new repos, it's always messy, sorry about all this. I also dislike the choose-your-own-adventure way we have to write docs to map out a path for users :-(.

To put it in perspective, PDB3 only supports Puppet 4 for packaging reasons. You need to be ready to move into that modern world to get both up to date.

earsdown:
@kbarber fair enough.
@rameshmimit I solved this for me by throwing the following fact into my profile module's lib/facter directory, then setting $::puppetdb::globals::version to $::puppetdb_version ... 
```
Facter.add(:puppetdb_version) do
    setcode do
        osfamily = Facter.value('osfamily')
        case osfamily
        when "Debian"
            command='/usr/bin/apt-cache show --no-all-versions puppetdb | awk " \$1 == \"Version:\" { print \$2 } "'
            version = Facter::Util::Resolution.exec(command)
            if version =~ /^(.+)$/
                $1
            else
                nil
            end
        when "RedHat", "Suse"
            command='/bin/yum -d 0 -e 0 -y list puppetdb 2> /dev/null | awk " \$1 == \"puppetdb.noarch\" { print \$2 } "'
            version = Facter::Util::Resolution.exec(command)
            if version =~ /^(.+)$/
                $1
            else
                nil
            end
        else
            nil
        end
    end
end
```

Hope this helps.

PR #207, opened 2015-09-04
MODULES-2556 add postgres_datadir parameter

westbywest:
This PR includes parametrization of the PostgreSQL datadir for PuppetDB (which otherwise just uses the puppetlabs-postgresql module default).  This allows locating PostgreSQL data outside of the host's root volume for performance advantage, i.e. on a faster AWS volume or hard disk with higher RPM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,10 @@ Conditionally manages the PostgresQL server via `postgresql::server`. Defaults t
 
 The URL to use for testing if the PuppetDB instance is running. Defaults to `/pdb/meta/v1/version`.
 
+####`postgres_datadir`
+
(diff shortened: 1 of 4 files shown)


bmjen:
@westbywest Thanks for the contribution! I think this LGTM, however with the new exposed parameter, can you write a test for it?

kbarber:
I'm wondering if this is the right move, one could potentially end up proxying all postgresql fields here if you took this pattern to its logical conclusion. I wonder if instead we should be promoting the usage of postgresql separately, when extra control is needed like this.

@westbywest this PR is 24 days old, which makes it a little stale, please re-open once you've addressed the comments to the PR.

PR #209, opened 2015-09-15
(PDB-1913) manage vardir

wkalt:
This updates the module to manage vardir, and also makes room for general
management of the global ini section.


manifests/params.pp
```diff
@@ -115,6 +118,7 @@
     }
     $terminus_package       = 'puppetdb-termini'
     $test_url               = '/pdb/meta/v1/version'
+    $vardir                 = '/opt/puppetlabs/server/data/puppetdb'
     $database_embedded_path = '/opt/puppetlabs/server/data/puppetdb/db/db'
```
ajroetker: This path should be getting it's value from vardir, `$database_embedded_path = "${vardir}/...`

ajroetker:
LGTM gonna test an install now

PR #210, opened 2015-09-21
(PDB-1430) overwritable java_args

wkalt:
this creates a new config setting, merge_default_java_args, that users may set
to false to override the default java_args.


ajroetker:
@kbarber this looks good to me but would you mind taking a look as well?

wkalt:
I learned today that this won't work on versions of puppetdb core that package with ezbake up to and including the latest version, since ezbake itself hardcodes some default java args. This won't affect the 2.x version that motivated this patch, and shouldn't affect 3.x versions in the near future. Shouldn't prevent this being merged imo.

PR #211, opened 2015-10-05
(maint) Lint fixes

kbarber:
This patch fixes some of the lint warnings/errors we've been getting on the
forge.

In some cases I've fixed real bugs also, in particular double quotes/single
quotes in cases, which would have broken variable interpolation.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -72,7 +72,7 @@ class puppetdb::params inherits puppetdb::globals {
       'RedHat', 'Suse', 'Archlinux','Debian': {
         $confdir                = '/etc/puppetdb/conf.d'
         $vardir                 = '/var/lib/puppetdb'
-        $database_embedded_path = '${vardir}/db/db'
+        $database_embedded_path = "${vardir}/db/db"
         $puppet_confdir         = pick($settings::confdir,'/etc/puppet')
         $puppet_service_name    = 'puppetmaster'
         $ssl_dir                = '/etc/puppetdb/ssl'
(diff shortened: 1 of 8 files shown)


manifests/params.pp
```diff
@@ -80,7 +80,7 @@
       'OpenBSD': {
         $confdir                = '/etc/puppetdb/conf.d'
         $vardir                 = '/var/db/puppetdb'
-        $database_embedded_path = '${vardir}/db/db'
+        $database_embedded_path = "${vardir}/db/db"
```
ajroetker: oh wow haha nice catch

manifests/params.pp
```diff
@@ -88,50 +88,59 @@
...
     $terminus_package = 'puppetdb-terminus'
     $test_url         = '/v3/version'
   } else {
     case $::osfamily {
       'RedHat', 'Suse', 'Archlinux','Debian': {
-        $confdir                = '/etc/puppetlabs/puppetdb/conf.d'
-        $puppet_confdir         = pick($settings::confdir,'/etc/puppetlabs/puppet')
-        $puppet_service_name    = 'puppetserver'
-        $ssl_dir                = '/etc/puppetlabs/puppetdb/ssl'
+        $confdir             = '/etc/puppetlabs/puppetdb/conf.d'
```
ajroetker: These should probably be `"` double quotes if we're fixing up syntax.

ajroetker: Sorry scratch that, got confused

PR #212, opened 2015-10-09
fix legacy_terminus value for puppetdb::master::puppetdb_conf class

cristifalcas:


ajroetker:
Could you explain what issue this is fixing @cristifalcas? The intention of drawing that value from `$puppetdb::params` was so that someone who uses a different package name for the old terminus still gets the correct terminus configuration.

cristifalcas:
The avalue was always false.

$puppetdb::params is not configurable. Maybe $puppetdb::params::terminus_package should have been used? But $puppetdb class is not used on all masters, instead puppet:master:config should be used.

Also, the class accepts as a parameter a variable $terminus_package that can and should be used?

ajroetker:
The value isn't necessarily always `false`, if you set 
```ruby
    class {'puppetdb::globals':
        version => '2.3.5-1.el7',
    }
```
in your manifest, the conditional https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/params.pp#L70 here sets https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/params.pp#L70 the terminus name to puppetdb-terminus otherwise we use the PuppetDB 3.0.0 default https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/params.pp#L126 puppetdb-termini. This setting only should change when someone changes the PuppetDB version being managed to something less than 3.0.0, not based on any user supplied value for the terminus package name.

cristifalcas:
We have multiple puppet masters and puppet db is only installed on 1 of them. So, on the other masters we use only the class puppetdb::master::config and we don't call class puppetdb anywhere. Because of this, the value is always false in our case (it's true only on the machine where we also define the puppetdb class).

I hope it's more clear now.

ajroetker:
@cristifalcas Ah I see your issue. So there are two solutions for your problem depending on your setup.

The first option would be if all your masters are managed via a single site.pp, where you would just put the `puppetdb::globals` class declaration with `version => <your version>` at the top level of the site.pp, a global variable for all your nodes to use.

The second option would be to declare
```
class {'puppetdb::globals':
  version => '2.3.5-1.el7',
}
```
Next to your declarations of the `puppetdb::master::config` class.

cristifalcas:
Yes, but why bother? And, anyway, I don't see how your examples are fixing this, considering that the comparison is with $puppetdb::params::terminus_package, which is static and not with $puppetdb::terminus_package, which can be overwritten?

$terminus_package already defaults to $puppetdb::params::terminus_package, which is the same string as in 

    legacy_terminus    => $puppetdb::params::terminus_package == 'puppetdb-terminus',

And can be overwritten when calling the class. What's the problem with using the available parameter?

Also, I don't see how is not obvious that comparing constants is a mistake and will always have the same result. The constants here being $puppetdb::params::terminus_package and 'puppetdb-terminus'.

ajroetker:
Let me go back and try and paint the full picture in a single comment.

The parameter `legacy_terminus` of the `puppetdb::master::puppetdb_conf` class is used to determine whether the module should use the PDB 2.x terminus config format (when `legacy_terminus == true`) or the PDB 3.x terminus config format (when `legacy_terminus == false`). This parameter is not intended to be exposed to the user, the intent is for this variable to be set depending on what version the user specifies in the `version` param of the `puppetdb::globals` class.

The parameter `terminus_package` of the `$puppetdb::master::config` class is used for specifying the name of the terminus package, which for most users will be `puppetdb-terminus` for PDB 2.x or `puppetdb-termini` for PDB 3.x. Some users package the terminus themselves though and might have different package names, such as FreeBSD users. Therefore the `terminus_package` value that a user supplies to this class is independent of whether the user is using PDB 2.x or 3.x (which is what we care about for the `legacy_terminus` value).

The parameter `$puppetdb::params::terminus_package` is *not* static. It's value is set depending on what version is set in the `puppetdb::globals` class (similar to the way the `puppetlabs-postgresql` module has a dynamic `postgresql::params` class which depends on the values set in `postgresql::globals`). [Here is the conditional in the params class](https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/params.pp#L70). We use the `$puppetdb::params::terminus_package == 'puppetdb-terminus'` here to determine what version of PDB and thus the terminus the user is using, rather than what the package name of the terminus is.

cristifalcas:
Sorry about that, I didn't noticed that the class calls the global class.

From your explanation it looks like the $terminus_package should not be a parameter and should not be used for installing the relevant package either?

    package { $terminus_package:
    }

The solution still doesn't work cleanly for us, as we have a specific profile for puppetdb and other profile for puppet masters. Incidentally, for one machine we use both of them, so we get a duplicate resource.

I will try to us "include puppetdb::globals" and set the parameter version from hiera. This should work for both roles.

PR #214, opened 2015-10-13
Add option to disable cleartext HTTP port

michaelweiser:


kbarber:
@michaelweiser looking good how about an rspec test for setting this, and not enabling it, for kicks? https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/spec/unit/classes/server/jetty_ini_spec.rb

The defaults area should be enough to test it disabled (so nothing to do here?), you'll need a new block for testing it when switched on.

michaelweiser:
@kbarber: Here you go.

PR #218, opened 2016-02-08
lint fix in report_processor

holtwilkins:


wkalt:
@holtwilkins thanks for the cleanups!

PR #220, opened 2016-02-08
(maint) use 'puppetdb-terminus' on redhat with version 3+

wkalt:
This makes it possible to upgrade PDB from 2.x to 3+ on redhat without
explicitly specifying the terminus package name.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -99,26 +99,34 @@ class puppetdb::params inherits puppetdb::globals {
     $test_url         = '/v3/version'
   } else {
     case $::osfamily {
-      'RedHat', 'Suse', 'Archlinux','Debian': {
+      'RedHat', 'Suse': {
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -99,26 +99,34 @@
     $test_url         = '/v3/version'
   } else {
     case $::osfamily {
-      'RedHat', 'Suse', 'Archlinux','Debian': {
+      'RedHat', 'Suse': {
+        $etcdir              = '/etc/puppetlabs/puppetdb'
+        $puppet_confdir      = pick($settings::confdir,'/etc/puppetlabs/puppet')
+        $puppet_service_name = 'puppetserver'
+        $terminus_package    = 'puppetdb-terminus'
```
ajroetker: Could you add a spec test for this?

PR #224, opened 2016-03-21
open_ssl_listen_port defaults to undef, not true

mmckinst:
changed in commit 448f8bc996b4e797ef463b1c2b89710374d9fd19


wkalt:
@mmckinst thanks!

PR #226, opened 2016-04-08
(MODULES-1171) pl-apt is not used

hunner:


kbarber:
We use postgresql and their upstream repos ... which uses apt ... doesn't it?

hunner:
You're right, so apparently it is needed.

PR #228, opened 2016-04-28
(PDB-2571) Ensure puppetdb.ini file has correct permissions

kbarber:
Much like read-database.ini, we need to ensure the permissions for puppetdb.ini are set explicitly
to ensure permissions are still correct after configuration. Without this users with different umask
settings may find their files are no longer accessible after the module runs.

### DIFF
--- a/manifests/server/command_processing.pp
+++ b/manifests/server/command_processing.pp
@@ -6,12 +6,6 @@ class puppetdb::server::command_processing (
   $confdir         = $puppetdb::params::confdir,
 ) inherits puppetdb::params {
 
-
-  file { "${confdir}/config.ini":
-    ensure => 'present',
-    mode   => '0644',
-  }
-
   # Set the defaults
   Ini_setting {
     path    => "${confdir}/config.ini",
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,23 @@
+# Final publish location for module build
 pkg/
+
+# VIM files
 *.swp
+
+# Mac files
 .DS_Store
+
+# Generated directory for coverage results
 coverage/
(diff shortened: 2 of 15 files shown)


ajroetker:
@kbarber Don't database.ini and jetty.ini have this same problem? Any reason not to fix those too?

kbarber:
@ajroetker actually they don't, they get installed by the package and are set absolutely to the correct permission afaik. This file resource I've fixed is only created by the module. Of course, we can always define permissions over all files we touch in the module if we think its a good idea, it may mean changing things in two places later on if we ever change our minds.

senior:
@kbarber any idea why there are ruby syntax errors on the older ruby versions we're testing?

kbarber:
@senior trailing commas in 1.8.7 aren't supported, my fault.

kbarber:
@ajroetker okay, I've been a little more ambitious now, and managed all the files. Found a bug also, and added tests where needed.

ajroetker:
Awesome, testing this now!

manifests/server/command_processing.pp
```diff
@@ -6,12 +6,6 @@
   $confdir         = $puppetdb::params::confdir,
 ) inherits puppetdb::params {
 
-
-  file { "${confdir}/config.ini":
```
kbarber: This just moved to global.pp, because well, it didn't feel like it should be here? I don't know - maybe we should just collapse both these pp files into one.

There are a lot of refactoring possibilities around configuration I think, right now we just keep copying files ... perhaps we can refactor it all into one big file, seems like its scaffolding for scaffolding sake.

PR #229, opened 2016-05-02
Manage the maximum-pool-size configuration parameter in database.ini

nward:
Tested with postgres in my environment, and works great.

Tests are added following the method of testing for other config parameters.


ajroetker:
👍

manifests/server/database.pp
```diff
@@ -147,4 +148,9 @@
     setting => 'conn-lifetime',
     value   => $conn_lifetime,
   }
+
+  ini_setting { 'puppetdb_database_max_pool_size':
+    setting => 'maximum-pool-size',
```
ajroetker: This setting is actually called `partition-conn-max` on PRE 4.0 PDB

kbarber: This should be set at least optionally, and not be set when the module has no setting defining it. Also, we are baking in the setting of 25 into the module, we should just undef it, and only set when the user has set it explicitly.

Unless we ant to go and do all the work for making this backwards compatible also, the duplicated default-in-module still exists though.

kbarber:
@nward as this stands, this will break 3.0 users, so we need to make the setting of this parameter to be only upon user choice, so if the user defines it, set it, otherwise do not set it.

kbarber:
Hi @nward, have you had time to address the concerns in this PR review yet?

ajroetker:
Looks like this was added onto in https://github.com/puppetlabs/puppetlabs-puppetdb/pull/232, closing in favor of the aforementioned PR :) thanks for the contribution!

nward:
Awesome, sorry, crazy few months so didn't get a chance to sort this out myself. Glad there's another fix!

PR #230, opened 2016-05-02
Add a manage_db parameter to let the user the freedom to manage

buzzdeee:
the postgresql database creation outside of the puppetdb module.

An other option would be to include the functionality within the
manage_dbserver parameter, if set to false/true, instead of introducing
another parameter.

Let me know if you would be in favor of the manage_dbserver parameter
takes care of it, and I change it, or if you prefer, what I did so far.

Then I can change if necessary, and add some tests for it.

cheers,

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,11 @@ Specify the supported SSL protocols for PuppetDB (e.g. TLSv1, TLSv1.1, TLSv1.2.)
 
 If true, the PostgreSQL server will be managed by this module. Defaults to `true`.
 
+###`manage_db`
+
+If true, the PostgreSQL database in the PostgreSQL instance will be managed by
+this module. Defaults to `true`.
+
 ####`database`
 
 Which database backend to use; legal values are `postgres` (default) or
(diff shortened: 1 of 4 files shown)


buzzdeee:
for your convenience, rebased to current master

ajroetker:
@buzzdeee Thanks for the contribution!

Are you wanting this change so that you can use the `class { 'puppetdb': }` without the module managing anything PostgreSQL related? If so, we support that using `class { 'puppetdb::server': }`.

Or are you wanting to have `class { 'puppetdb': manage_dbserver => true, manage_db => false }`?

buzzdeee:
yes, the first case it what I'm trying to get to. What I currently have is just an 
'include puppetdb' in my profile, and I don't want that to touch any db related stuff, only configuring itself.
I wasn't aware that I could just change my profile to include puppetdb::server to get away with what
I want. So I just changed my setup to use the puppetdb::server, and it seems to work as advertized,
and is something I can live with.

Because of that, I don't mind if you close this PR without merging.

ajroetker:
Thanks again for the contribution @buzzdeee I'm glad `include puppetdb::server` worked for you as advertised :) feel free to raise another PR if that ends up not working for you.

PR #232, opened 2016-05-29
Manage the pool size configuration parameters in database.ini

kpaulisse:
https://github.com/puppetlabs/puppetlabs-puppetdb/pull/229 is a PR from a few weeks ago to add pool size configuration parameters to database.ini. I also needed this feature, but I run PuppetDB 3.2 and because of differences in the parameter name, I needed to make some updates. Unfortunately, I didn't receive a response when I submitted these changes to the author of that PR (in https://github.com/nward/puppetlabs-puppetdb/pull/1) so I am opening this directly here.

In this implementation, the default is that `partition-conn-max`/ `maximum-pool-size` will not be put into the configuration files. If specified by appropriately named hiera keys or other lookup method, then `partition-conn-max` will be configured for PuppetDB 3.2, and `maximum-pool-size` will be configured for PuppetDB 4.0 and later.

I have tested this on PuppetDB 3.2 and it works. I don't have PuppetDB 4.0 so I can't test the end result there. I've also resolved the merge conflicts between https://github.com/puppetlabs/puppetlabs-puppetdb/pull/229 and the current master, to make this merge-able.


ajroetker:
@kpaulisse Thanks for the contribution! I'll test this out.

manifests/params.pp
```diff
@@ -164,4 +166,13 @@
   $certificate_whitelist      = [ ]
   # change to this to only allow access by the puppet master by default:
   #$certificate_whitelist      = [ $::servername ]
+
+  # Get the parameter name for the database connection pool tuning
+  if $puppetdb_version in ['latest','present'] or versioncmp($puppetdb_version, '4.0.0') >= 0 {
+    $database_max_pool_size_setting_name = 'maximum-pool-size'
+  } elsif versioncmp($puppetdb_version, '3.2.0') >= 0 {
```
ajroetker: We could set this all the way to '2.8.0' at least!

kpaulisse: I'm absolutely happy to do that - but I have no way of testing anything less than 3.2 for myself.

ajroetker: No problem I've got you covered there, I've verified this on 2.8.7 so it should be ok for the whole of the 2.8 series (which is as far back as we support anyways).

kpaulisse: Updated code to change 3.2.0 to 2.8.0 for that comparison. Thanks!

ajroetker:
👍 from me. Will merge when I get another quick sanity check from another PuppetDB team member.

PR #235, opened 2016-06-23
(PDB-2814) Add class for installing CLI

ajroetker:
This commit adds a class for installing and managing the PuppetDB CLI to
the PuppetDB module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -856,6 +856,49 @@ be installed from the regular repository. Defaults to `true`.
 If the postgresql.org repo is installed, you can install several versions of
 postgres. Defaults to `9.4`, the latest stable version.
 
+### puppetdb::cli
+
+The `puppetdb::cli` class manages the `puppet-client-tools` package which
+contains the PuppetDB CLI. This class also manages the associated PuppetDB CLI
+configuration file.
+
+You must declare the class to use it:
+
+    class { 'puppetdb::cli': }
+
+**Parameters within `puppetdb::cli`:**
+
+####`version`
+
+The version of the `puppet-client-tools` package that should be installed. You
+may specify an explicit version number, 'present', or 'latest' (defaults to
+'present').
+
+####`server_urls`
+
+The urls to use for contacting PuppetDB. This parameter may be either a String
+or an Array[String]. By default the value is `['http://localhost:8080']`.
+
+####`confdir`
+
+The directory in which to place the configuration file, `puppetdb.conf`, for the
+PuppetDB CLI. By default this path is the global configuration path at
+`/etc/puppetlabs/client-tools`. You may wish to change this to
+`$HOME/.puppetlabs/client-tools` to manage the per-user configuration file (when
+diverging from the global configuration).
+
+####`localcacert`
+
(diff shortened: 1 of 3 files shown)


ajroetker:
@kbarber @wkalt @senior @rbrw
I'm not sure what exactly this class should look like so I'm putting the PR up for review for now to start discussion.

With the current PR, a user would install the CLI like so

```
include puppetdb::cli
# or
# class { 'puppetdb::cli' : puppetdb_urls => ["https://#{puppetdb_host}:#{puppetdb_port}"] }
```

I have the defaults for the SSL credentials set to the agent certs, but the url set to http and localhost, should I default the SSL credentials at all if assuming the CLI will be installed on the PDB node by default?

Should we be defaulting to localhost for the puppetdb url? We could try and default to using SSL and checking if the puppetdb_host and puppetdb_port params are defined on the puppetdb class.

Should we attempt to validate version compatibility (i.e. <= PDB 3 don't support the CLI)?

If we do try to default to SSL by default for the CLI, should we attempt to warn users that their agent certs (or which certs we default to) need to be on the PuppetDB cert whitelist?

What I think makes the most sense...

Default to nothing for SSL credentials at all and use localhost for the urls by default, and document more advanced usage from there.

ajroetker:
We should wait to merge this until we've released the new version of the CLI.

manifests/cli.pp
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  file { $puppetdb_cli_config:
+    ensure  => 'present',
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0444',
+    content => inline_template('<%= @_puppetdb_cli_config.to_json %>'),
+    require => Package['puppet-client-tools'],
+  }
```
ajroetker: All this code is mostly lifted from the PE module (puppetlabs-puppet_enterprise).

README.md
```diff
@@ -856,6 +856,49 @@ be installed from the regular repository. Defaults to `true`.
...
+
+**Parameters within `puppetdb::cli`:**
+
+####`version`
+
+The version of the `puppet-client-tools` package that should be installed. You
+may specify an explicit version number, 'present', or 'latest' (defaults to
+'present').
+
+####`puppetdb_urls`
```
wkalt: why invent a new term here? is it going to conflict with server_urls somewhere else?

ajroetker: I lifted this CLI management from the PE module. We don't actually use `server_urls` anywhere in the module (i.e. we assume one PuppetDB host/port for everything else in the module.

ajroetker:
@wkalt I updated this finally for your comments.

wkalt:
@senior @ajroetker I'm thinking this may be blocked on CT-18. We don't have client tools packages for debian yet, and I've got reservations about introducing module features that are only available for certain supported operating systems.

senior:
@ajroetker sounded like we were waiting on a CLI release or something here? any update?

preisbeck:
Is there any chance that version 5.2.0 will hit the forge soon, or is there a patch released planed? I'd like to use some new parameter that are exclusive to the Git version.

PR #236, opened 2016-07-20
replace invalid node name variables in 3-node example

jason-azze-cdna:
The basic 3-node setup example in the documentation assigns some variables for the 3 nodes and then tries to resolve these variables in the node statements. However, puppet complains of a syntax error when it reaches the variable just after the `node` keyword.

`puppet parser validate` also indicates that code is not valid.

The puppet language reference says...
>A node statement’s name must be one of the following:
>    A quoted string containing only letters, numbers, underscores (_), hyphens (-), and periods (.).
>    A regular expression.
>    The bare word default.

Newbies (like I am) reading the examples might get confused (like I did). :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,14 +106,14 @@ scenario might look like:
     $master_host = 'master.example.lan'
     $puppetdb_host = 'puppetdb.example.lan'
     $postgres_host = 'postgres.example.lan'
-    node $master_host {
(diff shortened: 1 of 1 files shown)


wkalt:
@jason-azze-cdna thanks!

PR #239, opened 2016-08-26
Fixed arrow alignment and line lengths to pass lint checks

danieljamesscott:
Fixes some lint check warnings. There are others related to 'class inheriting from params'. But they'll require a much more significant change.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -70,11 +70,13 @@ class puppetdb::server (
 ) inherits puppetdb::params {
   # deprecation warnings
   if $database_ssl != undef {
-    warning('$database_ssl is deprecated and will be removed in the next major release. Please use $jdbc_ssl_properties = "?ssl=true" instead.')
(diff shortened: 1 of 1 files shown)


manifests/server.pp
```diff
@@ -70,11 +70,13 @@
 ) inherits puppetdb::params {
   # deprecation warnings
   if $database_ssl != undef {
-    warning('$database_ssl is deprecated and will be removed in the next major release. Please use $jdbc_ssl_properties = "?ssl=true" instead.')
+    warning('$database_ssl is deprecated and will be removed in the next major release.
```
kbarber: This mechanism, where you are breaking up the lines as such to achieve < 80 chars is going to mess with the output formatting of the warning.

See this for an example: https://gist.github.com/kbarber/1fb25ad0b2b70383fb3867a25b7fd451

kbarber:
Closing due to inactivity. Please amend and re-raise based on the comments provided.

PR #243, opened 2016-09-12
Add support for multiple puppetdb servers

cwood:
Simple change to allow a comma delimited list of puppetdb servers to choose from.

### DIFF
--- a/manifests/master/puppetdb_conf.pp
+++ b/manifests/master/puppetdb_conf.pp
@@ -3,6 +3,7 @@
 class puppetdb::master::puppetdb_conf (
   $server             = 'localhost',
   $port               = '8081',
+  $servers            = [],
   $soft_write_failure = $puppetdb::disable_ssl ? {
     true => true,
     default => false,
@@ -30,9 +31,16 @@ class puppetdb::master::puppetdb_conf (
       value   => $port,
     }
   } else {
-    ini_setting { 'puppetdbserver_urls':
-      setting => 'server_urls',
-      value   => "https://${server}:${port}/",
+    unless $servers {
+      ini_setting { 'puppetdbserver_urls':
+        setting => 'server_urls',
+        value   => "https://${server}:${port}/",
+      }
+    } else {
+      ini_setting {'puppetdbserver_urls':
+        setting => 'server_urls',
+        value   => join($servers, ',')
+      }
     }
   }
 
--- a/manifests/master/config.pp
+++ b/manifests/master/config.pp
@@ -12,6 +12,7 @@ class puppetdb::master::config (
     true    => $::puppetdb::disable_ssl,
     default => false,
   },
+  $puppetdb_servers            = [],
   $masterless                  = $puppetdb::params::masterless,
   $puppetdb_soft_write_failure = false,
   $manage_routes               = true,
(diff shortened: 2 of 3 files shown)


wkalt:
@cwood thanks, this looks good to me. @ajroetker any thoughts?

wkalt:
@cwood actually this is hitting spec test failures. spec tests are fixed now, so you should be able to get fixed locally.

manifests/master/puppetdb_conf.pp
```diff
@@ -3,6 +3,7 @@
 class puppetdb::master::puppetdb_conf (
   $server             = 'localhost',
   $port               = '8081',
+  $servers            = [],
```
ajroetker: I think this should probably be called `server_urls` to indicate that it need to be a list of urls rather than just hostnames.

manifests/master/puppetdb_conf.pp
```diff
@@ -30,9 +31,16 @@
...
+    unless $servers {
+      ini_setting { 'puppetdbserver_urls':
+        setting => 'server_urls',
+        value   => "https://${server}:${port}/",
+      }
+    } else {
+      ini_setting {'puppetdbserver_urls':
+        setting => 'server_urls',
+        value   => join($servers, ',')
+      }
```
ajroetker: What would you think about adding a sort here as well?

manifests/master/puppetdb_conf.pp
```diff
@@ -30,9 +31,16 @@
...
-    ini_setting { 'puppetdbserver_urls':
-      setting => 'server_urls',
-      value   => "https://${server}:${port}/",
+    unless $servers {
+      ini_setting { 'puppetdbserver_urls':
+        setting => 'server_urls',
+        value   => "https://${server}:${port}/",
+      }
+    } else {
+      ini_setting {'puppetdbserver_urls':
```
ajroetker: Could we put a warning here if `$server and $port` are set to something other than the default, i.e. someone tried to set both urls and those settings?

manifests/master/puppetdb_conf.pp
cwood: I mean sorting seems like it would enforce an order you may or may not want.I don't know for sure. But I would guess that puppet does them in the order you put them in. Maybe someone might want them in a specific order?

manifests/master/puppetdb_conf.pp
cwood: Yea I can do that.

manifests/master/puppetdb_conf.pp
cwood: Makes sense.

manifests/master/puppetdb_conf.pp
ajroetker: Yeah disregard this comment, order does matter here as the first PuppetDB in the list is the first to get queried :)

ajroetker:
Hi @cwood! Are you still working on this PR?

cwood:
hey @ajroetker sorry I haven't gotten back to this yet. I have implemented this so I would need to redo my own stuff to make this work. Hopefully soon.

mullr:
Closing due to inactivity

PR #244, opened 2016-10-13
(PDB-3060) Add concurrent-writes parameter.

aperiodic:
Add a parameter to govern the command-processing config section's
`concurrent-writes` setting.


aperiodic:
@kbarber sorry, this fell through the cracks. Updated w/README addition.

PR #245, opened 2016-10-20
Add postgres_datadir to params

petrenkoas83:
This modification allow change postgresql data_directory.
Example from nodes.pp:

```
node puppet-db {
  class { 'puppetdb::database::postgresql':
    listen_addresses => '0.0.0.0',
    postgres_datadir => '/media/data/postgresql/9.4/main',
  }
}
```

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -9,12 +9,14 @@ class puppetdb::database::postgresql(
   $manage_server        = $puppetdb::params::manage_dbserver,
   $manage_package_repo  = $puppetdb::params::manage_pg_repo,
   $postgres_version     = $puppetdb::params::postgres_version,
+  $postgres_datadir     = $puppetdb::params::postgres_datadir,
 ) inherits puppetdb::params {
 
   if $manage_server {
     class { '::postgresql::globals':
       manage_package_repo => $manage_package_repo,
       version             => $postgres_version,
+      datadir             => $postgres_datadir,
     }
     # get the pg server up and running
     class { '::postgresql::server':
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -12,8 +12,7 @@ class puppetdb::params inherits puppetdb::globals {
   $disable_ssl               = false
   $open_ssl_listen_port      = undef
   $postgres_listen_addresses = 'localhost'
-  $postgres_datadir          = '/var/lib/postgresql/9.4/main'
-
+  
   $puppetdb_version          = $puppetdb::globals::version
   $database                  = $puppetdb::globals::database
   $manage_dbserver           = true
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -12,6 +12,7 @@ class puppetdb::params inherits puppetdb::globals {
   $disable_ssl               = false
   $open_ssl_listen_port      = undef
   $postgres_listen_addresses = 'localhost'
+  $postgres_datadir          = '/var/lib/postgresql/9.4/main'
 
   $puppetdb_version          = $puppetdb::globals::version
   $database                  = $puppetdb::globals::database


manifests/params.pp
```diff
@@ -12,6 +12,7 @@
   $disable_ssl               = false
   $open_ssl_listen_port      = undef
   $postgres_listen_addresses = 'localhost'
+  $postgres_datadir          = '/var/lib/postgresql/9.4/main'
```
kbarber: We definitely should not be reflecting our own opinions for the default of this parameter over in the puppetdb module, those opinions should be expressed in postgresql only. Otherwise, we're just going to play a game of out-of-sync catch up over time, if upstream changes its values for example.

manifests/database/postgresql.pp
```diff
@@ -9,12 +9,14 @@
   $manage_server        = $puppetdb::params::manage_dbserver,
   $manage_package_repo  = $puppetdb::params::manage_pg_repo,
   $postgres_version     = $puppetdb::params::postgres_version,
+  $postgres_datadir     = $puppetdb::params::postgres_datadir,
```
kbarber: The thing about these kind of parameters that get proxied over to the postgresql module, is that you never win unless you do them all. They'll continue to be added and added by contributors until we have about 80% of them, never quite all of them.

I'm wary of adding any more of these to this module as really, if the functionality is required, one should be breaking out from this - and using the postgresql module independently instead. The other alternative is to provide a way to proxy everything (like passing a hash), without hard-coding defaults (like below) as well, so it acts as a really dumb proxy rather then having each parameter manually copied over.

But really, what is the value of having this as a convenience (when considering option 2) when really one can break out from using this class, and use postgresql directly instead.

manifests/params.pp
petrenkoas83: I cannot disagree by this statement.

kbarber:
@petrenkoas83 I think the tests are now failing with your recent change. What are your thoughts about the other comment that I've left on this ticket, in a way, I'm almost against even having this feature added at all to the module, but I'm happy to be convinced otherwise.

manifests/database/postgresql.pp
petrenkoas83: I have not found another way to change the postgresql datadir in case if i use puppetlabs-puppetdb module.
If i use class puppetdb, as stated in the [documentation](https://docs.puppet.com/puppetdb/4.3/install_via_module.html#step-2-assign-classes-to-nodes)
```
class { 'puppetdb':
    listen_address => 'example.foo.com'
}
```

, and i trying to use class postgresql, i get error "cannot redeclare class" or something like that. At this moment I am not at my workplace, and i can not provide the exact text of the error.

kbarber: Its because class `puppetdb` also configures postgresql, but if instead you use `puppetdb::server` it decomposes `puppetdb` and doesn't pull in postgresql:

```
class { 'puppetdb::server':
  listen_address => '0.0.0.0',
}

class { ':postgresql::globals':
  manage_package_report => true,
  version => '9.4',
}

class { '::postgresql::server':
  datadir => '/srv/pg_data',
}

  database  => 'puppetdb',
}

postgresql::server::db { $database_name:
  user     => 'puppetdb',
  password => 'mypassword',
  grant    => 'all',
}
```

Another decomposition option, is to pass in manage_server => false, which doesn't manage postgresql, but creates the database and user for you:

```
class { 'puppetdb':
  listen_address => '0.0.0.0',
  manager_dbserver => false,
}

class { ':postgresql::globals':
  manage_package_report => true,
  version => '9.4',
}

class { '::postgresql::server':
  datadir => '/srv/pg_data',
}

  database  => 'puppetdb',
}
```

petrenkoas83: Thank you for advice, Ken. I close my pull request.

kbarber: @petrenkoas83 no worries. If you feel you can find an alternate approach that improves this experience, feel free to raise a PR or whatever :-).

PR #246, opened 2016-11-03
Add support for Ruby 2.3.1

ghoneycutt:
The Puppet AIO is going to start using ruby 2.3.1 and as such needs to
be supported here.


ghoneycutt:
Thanks Ken!

ghoneycutt:
Can you please cut a new release?

PR #247, opened 2016-12-13
add option to customize cipher suites in jetty

selyx:
I've added an option to configure the [cipher-suites](https://docs.puppet.com/puppetdb/latest/configure.html#cipher-suites) parameter in jetty.ini since we've had the need to use specific ciphers only.


selyx:
Is something preventing this PR from being merged?

PR #248, opened 2017-01-19
module always requires a local postgresql

attachmentgenie:
Currently this module still requires a local postgresql server even when connecting to a remote one. As it always tries to manage a local postgresql db, this change fixes that problem.

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -27,12 +27,12 @@ class puppetdb::database::postgresql(
     postgresql::server::extension { 'pg_trgm':
       database  => $database_name,
     }
-  }
-
-  # create the puppetdb database
-  postgresql::server::db { $database_name:
-    user     => $database_username,
-    password => $database_password,
-    grant    => 'all',
+  
+    # create the puppetdb database
+    postgresql::server::db { $database_name:
+      user     => $database_username,
+      password => $database_password,
+      grant    => 'all',
+    }
   }
 }


chrisowensboston:
I have a need for this and would appreciate a merge into master.... Thanks

ajroetker:
Hi @attachmentgenie and @chrisowensboston, thanks for raising this PR and the feedback respectively. This issue has been raised against the module before and I think what would fix both your problems, if I understand correctly, would be to use the `puppetdb::server` [class here](https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/README.md#puppetdbserver) rather than using the `puppetdb` class with `manage_server => false`. Looking at the documentation in the README this fact is definitely not as clear as it could be. I'll add some more info to the README and maybe a comment, into the class you're proposing an edit to, which states this. Let me know if using the `puppetdb::server` class directly would solve your problem. Thanks again!

chrisowensboston:
Andrew you are correct. In fact, there is a use case that the proposed change would beak: there is a postgresql db server whose overall configuration I don’t control, but I want to create the puppetdb database on it.  In that case I would want to declare (on the postgresql server) puppetdb::database::postgresql with manage:server => false.  I take back my request to accept this PR.

attachmentgenie:
@ajroetker thanks for your comments. Yes ::puppetdb::server works for my usage.

However i still feel that this request is valid as the use case that @chrisowensboston proposes feel weird/incomplete to me. 

Why manage only some parts of your database service? Either the database service is important as whole and manage it. Or its not (or managed elsewhere) and dont manage any of it in this module

chrisowensboston:
@attachmentgenie Consider the case where you obtain your database service from a third party, i.e. what you control is a database instance but not the postgresql server itself. You have the rights to create, update, delete databases, roles, etc., but you have no access to any of the server configuration.

attachmentgenie:
@chrisowensboston IMHO that seems very unlikely and alien scenario for a singe node setup. 

I know of no db-administrator that is happy to give you no admin access to the db server, but does accept you have full root access on the underlying OS. (which then allows you to change the server configuration).

Also having , from the point of puppet, a non managed entity on one of your most vital nodes that you do rely on heavily is a situation so unwanted that its not something you should provide for over other more valid/realistic use cases.

chrisowensboston:
@attachmentgenie consider the case of an AWS RDS database instance. I have full postgresql admin rights, i.e., I can create, delete, etc databases, roles, and the like but I have absolutely no access to the underlying OS; I don't even know what the underlying OS is.  Puppet will quite happily create a database on that server by sending it sql commands through a database connection (see ::postgresql::server::db), but I don't own the node. 

So if I wanted to set up a puppetdb node that used an AWS RDB instance as its back end database, I would want to be able to declare puppetdb::database::postgresql with manage_server => false.  Then puppet could create the database for me.  This behavior would go away if the ::postgresql::server::db block were moved inside the (if $manage_server) block.

attachmentgenie:
@chrisowensboston  as far as i can workout postgresql::server::db can only create a local db not remote ones

chrisowensboston:
@attachmentgenie It's a little convoluted, but... 

postgresql::server::db declares postgresql::server::database

postgresql::server::database declares defaults for Postgresql_psql (line 35) including path, connection settings, etc.  and then, (line 76) executes the SQL code by calling the function postgresql_psql, which is defined in lib/puppet/provider/postgresql_psql/ruby.rb. This command just invokes the psql command, which does not require that the server be local.

wkalt:
I'm closing this PR. The behavior is supported by the module, as discussed above.

chrisowensboston:
wkalt I'm not understanding what you meay by the behavior being supported. 

How would I create a database on an AWS RDS instance?

wkalt:
@chrisowensboston you said early on that your problems were solved by @ajroetker 's suggestion. I'm not following whether this RDS discussion is just a hypothetical, or whether it is your situation (meaning your problem is not solved), but my impression is we've diverged from the original PR.

If you are need to manage RDS and are unable to with the module, I think you can do it by using manage_server = false and managing the RDS instance with the postgres module. RDS drove https://github.com/puppetlabs/puppetlabs-postgresql/pull/688, so it should be supported.

If we need to add RDS support to the PuppetDB module (which I think would be reasonable if it isn't supported) that would be worth a ticket or PR that addresses RDS specifically. As you've noted, remote databases are supported, so the premise of the PR as it stands today is not really valid in my mind.

chrisowensboston:
I concur with closing this PR. It’s going to be a bit before I can test the suggested approach; I’ll get back to you with a report. Thanks.

PR #250, opened 2017-02-25
Fix duplicate resource errors for puppet service due to parse order

kpaulisse:
The `defined(Service[$puppet_service_name])` pattern in this module is parse-order dependent, and I was unable to declare a `service { 'puppetserver' ...}` in some code I was writing because this module was parsed first and had already declared the resource.

To work around this, I've added a parameter to the class, which defaults to the existing behavior of `true` so it won't break existing behavior for anyone. However, one can override this to `false` which skips the creation of the `$puppet_service_name` service resource in this module and thereby avoids a duplicate resource error when compiling the catalog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -767,6 +767,14 @@ If `true`, the module will store values from `puppetdb_server` and `puppetdb_por
 parameters in the PuppetDB configuration file. If `false`, an existing PuppetDB
 configuration file will be used to retrieve server and port values.
 
+####`create_puppet_service_resource`
+
+If `true`, AND if `restart_puppet` is true, then the module will create a service
+resource for `puppet_service_name` if it has not been defined. Defaults to `true`.
+If you are already declaring the `puppet_service_name` service resource in another
+part of your code, setting this to `false` will avoid creation of that service
+resource by this module, avoiding potential duplicate resource errors.
+
 ####`strict_validation`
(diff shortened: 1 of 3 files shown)


mullr:
Hey @kpaulisse , thanks for the PR!

Two comments:
- First, by way of consistency, what do you think about changing this param to be named `manage_puppet_service` or something similar?
- Second, could we ask for a test case (probably in spec/unit/classes/master/config_spec.rb) to exercise this? It's clearly simple code, but it'll help us keep it working in the future. 

Thanks!

kpaulisse:
@mullr I'll look at adding a test.

My personal opinion is that `manage_puppet_service` is not an appropriate name, because this code is really not managing the service. Rather it's creating the resource if it doesn't exist yet in the portion of the catalog that's been evaluated thus far. While I understand why you're doing this, subtle changes can suddenly cause this to break - in my particular case it was the simple rename of a class that caused that things to be evaluated in a different order and therefore start throwing a duplicate resource error. So I personally believe that "create puppet service resource" is exactly what this is doing, but it's not really managing anything. However I'll defer to your wishes regarding the variable name, so if you want `manage_puppet_service` or something else, even in light of this explanation, tell me what you want it to be named and I'll make it happen.

PR #251, opened 2017-03-09
replace validate_* methods with datatypes

bastelfreak:
The goal of this is to get rid of most of the validate_* function calls since they are now deprecated.


mmoll:
this needs a rebase :)

bastelfreak:
Why is this still open :(

metadata.json
```diff
@@ -67,26 +67,26 @@
     "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.7.0"
+      "version_requirement": ">= 4.7.1 < 5.0.0"
```
eputnam: bump to < 6.0.0 plz

eputnam: just kidding

bastelfreak: tested it locally, doesnt work, reverted the commit. Fixing the issues for puppet5 should be handled in another PR.

PR #254, opened 2017-04-20
(PDB-3318) Better defaults for node-ttl, node-purge-ttl

mullr:
Default node-ttl to 7 days and node-purge-ttl to 14 days. This helps keep cruft
from piling up in the database in the default configuration, without user
intervention.


senior:
@mullr @ajroetker I've not seen PR tests for the module before, but they're failing. Any idea what's going on?

PR #255, opened 2017-04-25
(maint) Require Puppet 4.7 or greater

mullr:
Modules we depend on require this now.


senior:
Doesn't this require a major version bump?

PR #257, opened 2017-04-26
Add disable-update-checking parameter

baurmatt:
This PR adds sort for enabling disable-update-checking.


baurmatt:
Failing tests isn't related to this PR.

PR #258, opened 2017-05-15
Add options to set the source of the ssl certs

edestecd:
Also add a convenient boolean to copy over the certs from the puppet cert directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 Release notes for the puppetlabs-puppetdb module.
 
+#### 7.6.0 - 2020/09/02
+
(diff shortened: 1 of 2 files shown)


edestecd:
The master branch of the apt module is breaking on puppet 3, with a syntax error.

edestecd:
Also pinned modules in fixtures until this module depreciates puppet 4. This gets the tests passing again...

mullr:
Well, this PR was a long time ago, wasn't it. Sorry about the very, very long delay. 

Basically it looks good to me, but it needs tests. Can you add them and rebase onto master? Shouldn't need the module pinning anymore either.

PR #259, opened 2017-06-23
MODULES-5123 Fixed compatibility w/ puppetlabs-postgresql 5.0.0

jkroepke:
https://tickets.puppetlabs.com/browse/MODULES-5123

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -22,7 +22,7 @@ class puppetdb::params inherits puppetdb::globals {
 
   # The remaining database settings are not used for an embedded database
   $database_host          = 'localhost'
-  $database_port          = '5432'
+  $database_port          = 5432
   $database_name          = 'puppetdb'
   $database_username      = 'puppetdb'
   $database_password      = 'puppetdb'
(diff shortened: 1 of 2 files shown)


baurmatt:
`database_port` also needs to be a Integer.

jkroepke:
@baurmatt It's not necessary, because the database_port param does't have a validator

https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/validate_db_connection.pp#L11

But if you want it, it's okay.

baurmatt:
True, but https://github.com/puppetlabs/puppetlabs-postgresql/blob/5.0.0/manifests/server/grant.pp#L10 needs has a validation :(

jkroepke:
@baurmatt Maybe I do not hit this error because I updated a already working puppetdb.

Thanks for the hints.

jkroepke:
Hum, looks like test failures are not related to this issue..

mmoll:
this should also get noted in metadata.json then, 👍 otherwise

jkroepke:
Sure.

mullr:
Thanks for this PR! I ended up merging a different one that did the same thing, but appreciate the contribution.

PR #263, opened 2017-06-30
Move ntp dependency to spec_helper_acceptance

ekohl:
aac71812f6da18e402c8082d9627a37152943c75 added this dependency but it's only needed in the acceptance tests.


ekohl:
I'll try that.

ekohl:
Looks like the acceptance tests passed without the NTP module.

PR #266, opened 2017-07-09
README: Fixes for postgres default version and header rendering

jvperrin:
This fixes the rendering of the headers in the README on GitHub, and also fixes the PostgreSQL version mentioned since the default is now 9.6, not 9.4.


mullr:
lgtm, thanks for the PR!

PR #267, opened 2017-07-14
Fixing database_port variable to be an integer

RobReus:
This will fix the following error from puppetlabs-postgresql 5.0.0
`Error: Evaluation Error: Error while evaluating a Resource Statement, Postgresql::Server::Grant[database:GRANT puppetdb - all - puppetdb]: parameter 'port' expects an Integer value, got String`


RobReus:
Not having the error which appear in travis on my local system. Will dive into this later.

mmoll:
AFAICT this got fixed in the meanwhile...

RobReus:
@mmoll looks like it did. Closing PR.

PR #269, opened 2017-07-27
If we aren't managing the server, don't try to create dbs.

jadestorm:
This simply scoots the postgresql database creation up under the block where we are managing postgresql.  In my situation, I already have a postgresql db configured on my centralized postgresql server and don't need or want puppet to try to configure it.  (I also don't think the existing code would work anyway if you have manage_server turned off)

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -27,12 +27,12 @@ class puppetdb::database::postgresql(
     postgresql::server::extension { 'pg_trgm':
       database  => $database_name,
     }
-  }
 
-  # create the puppetdb database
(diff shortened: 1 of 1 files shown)


wkalt:
@jadestorm we previously closed this PR here: https://github.com/puppetlabs/puppetlabs-puppetdb/pull/248 but maybe this proves it's what the people want :)

Would you mind reviewing the discussion there to see if any of it addresses your concern? I need to refresh my memory myself.

jadestorm:
Ha!  Actually yes, that work work just fine for my usage.  =)  Thanks and closing the PR.  =)

PR #271, opened 2017-09-25
Error adding extension to PostgresSQL database with custom port

qingbo:
Seeing this error with custom port:

```
Error: /Stage[main]/Puppetdb::Database::Postgresql/Postgresql::Server::Extension[pg_trgm]/Postgresql_psql[Add pg_trgm extension to puppetdb]: Could not evaluate: Error evaluating 'unless' clause, returned pid 26843 exit 2: 'psql: FATAL:  database "puppetdb" does not exist
'
```

OS: CentOS 7.3.1611
Module versions:
mod 'puppetlabs/puppetdb', '6.0.1'
mod 'puppetlabs/postgresql', '5.1.0'

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -25,7 +25,8 @@ class puppetdb::database::postgresql(
     # get the pg contrib to use pg_trgm extension
     class { '::postgresql::server::contrib': }
     postgresql::server::extension { 'pg_trgm':
-      database  => $database_name,
+      database         => $database_name,
+      connect_settings => { 'PGPORT' => $database_port, },
     }
   }
 


pgassmann:
The problem is something else:

The dependency to the database is missing. This should be explicitly added. `postgresql::server::extension` does not add the dependency automatically, because the database is definded below, after the extension,  so the defined() check does not work. https://github.com/puppetlabs/puppetlabs-postgresql/blob/34aa0affa233e04a25c3b4d130e6db867bff08bd/manifests/server/extension.pp#L48-L50

```
      database  => $database_name,
      require   => Postgresql::Server::Db[$database_name],
    }
```

akomakom:
@pgassmann : it does work in a second run (although it seems that puppetdb fails to start unless you set disable_ssl to true and match listen_address with puppet::master::config (defaults don't match), but that's probably unrelated to this issue)

PR #272, opened 2017-10-18
add explicit dependency db -> extension.

pgassmann:
The dependency to the database is missing. This should be explicitly added. `postgresql::server::extension` does not add the dependency automatically, because the database is definded below, after the extension,  so the defined() check does not work. https://github.com/puppetlabs/puppetlabs-postgresql/blob/34aa0affa233e04a25c3b4d130e6db867bff08bd/manifests/server/extension.pp#L48-L50

```
      database  => $database_name,
      require   => Postgresql::Server::Db[$database_name],
    }
```
related to #271


Zak-Kent:
Thank you for adding this @pgassmann :) The test errors you saw will be fixed in an upcoming PR.

PR #276, opened 2018-01-19
Use puppet4 functions-api

juliantodt:
The legacy puppet3 functions-api should be avoided.
This migrates the existing functions to the new api and uses it in the module. It also adds tests for the new functions.


juliantodt:
The checks didn't fail locally as I was using the puppet5 version. Seems like puppet4 does some weird things when checking whether params are Hashes.

mmoll:
@juliantodt could you rebase this? Puppet 4.x was dropped from this module a while ago...

PR #277, opened 2018-03-05
add additional puppet-lint checks and fix errors

kpankonen:
We're using some additional puppet-lint checks (`absolute_classname` and `trailing_comma`) in our environment that were failing.  This MR adds those checks, and fixes the errors.


h0tw1r3:
lint issues addressed in #385

PR #278, opened 2018-06-01
Add support for DLO automatic cleanup

bastelfreak:


bastelfreak:
This is WIP. Let me know if you're interested in this and I will add tests + README.md updates.

Gemfile
```diff
@@ -2,10 +2,8 @@ source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
 group :development, :test do
   # Pinning `rake` because `v11` doesn't support Ruby 1.8.7
-  gem 'rake', '10.5.0'
+  gem 'rake'
   gem 'puppetlabs_spec_helper', :require => false
-  # Pinning due to bug in newer rspec with Ruby 1.8.7
-  gem 'rspec-core', '3.1.7'
```
bastelfreak: I run into some issues while running the tests locally. For debugging I removed the pinning.

austb: We shouldn't need to support such an old Ruby version anymore, so this should be ok.

manifests/params.pp
```diff
@@ -168,6 +168,7 @@
   $store_usage              = undef
   $temp_usage               = undef
   $disable_update_checking  = undef
+  $automatic_dlo_cleanup    = true
```
austb: We should default this to false and people can turn it on as needed for now. After it has been used a bit we can come back and look at defaulting to on.

templates/puppetdb-DLO-cleanup.timer.epp
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Daily Timer for puppetdb-DLO-cleanup
+
+[Timer]
+# Daily at 8am
+OnCalendar=*-*-* 08:0:00
```
austb: At a minimum, we should support customizing the time of day that this service runs. One could imagine wanting to schedule this for a low traffic time, which may be specific per user. It would also be nice to specific frequency. Some users may want to clean up once a week on Saturday.

Also the comment will no longer be accurate, so you can remove it or replace it with a general comment with a link to documentation on the [systemd timer syntax](https://www.freedesktop.org/software/systemd/man/systemd.time.html)

manifests/server.pp
```diff
@@ -297,6 +298,23 @@
...
+    if $facts['systemd'] {
+      # deploy a systemd timer + service to cleanup old reports
+      # https://puppet.com/docs/puppetdb/5.2/maintain_and_tune.html#clean-up-the-dead-letter-office
+      systemd::unit_file{'puppetdb-cleanup.service':
+        content => epp("${module_name}/puppetdb-DLO-cleanup.service.epp"),
+      }
+      -> systemd::unit_file{'puppetdb-cleanup.timer':
+        content => epp("${module_name}/puppetdb-DLO-cleanup.timer.epp"),
+      }
+      -> service{'puppetdb-cleanup.timer':
```
austb: This needs to be a `~>` so that the service will be restarted if the `puppetdb-cleanup.timer` file changes.

manifests/server.pp
```diff
@@ -297,6 +298,23 @@
     }
   }
 
+  if $automatic_dlo_cleanup {
+    if $facts['systemd'] {
+      # deploy a systemd timer + service to cleanup old reports
+      # https://puppet.com/docs/puppetdb/5.2/maintain_and_tune.html#clean-up-the-dead-letter-office
+      systemd::unit_file{'puppetdb-cleanup.service':
```
austb: We should rename this to `puppetdb-dlo-cleanup.service` (and the matching timer one below) in case we want to add another cleanup service in the future.

manifests/init.pp
```diff
@@ -148,6 +149,7 @@
     certificate_whitelist             => $certificate_whitelist,
     database_max_pool_size            => $database_max_pool_size,
     read_database_max_pool_size       => $read_database_max_pool_size,
+    automatic_dlo_cleanup             => $automatic_dlo_cleanup,
```
austb: We should add a configuration options here for time of day and then use it in the epp template below.

manifests/server.pp
```diff
@@ -297,6 +298,23 @@
...
+        content => epp("${module_name}/puppetdb-DLO-cleanup.service.epp"),
+      }
+      -> systemd::unit_file{'puppetdb-cleanup.timer':
+        content => epp("${module_name}/puppetdb-DLO-cleanup.timer.epp"),
+      }
+      -> service{'puppetdb-cleanup.timer':
+        ensure => 'running',
+        enable => true,
```
austb: I don't think this is necessary, but I could be mistaken.

manifests/server.pp
```diff
@@ -297,6 +298,23 @@
     }
   }
 
+  if $automatic_dlo_cleanup {
+    if $facts['systemd'] {
```
austb: If systemd does not exist, lets try setting up a cron job to clean this up instead. For example Redhat 6 has cron but not systemd.

templates/puppetdb-DLO-cleanup.timer.epp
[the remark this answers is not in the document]
austb: You could select the hour with `fqdn_rand(24)` and the minutes with `fqdn_rand(60)`.

austb:
@bastelfreak These comments are done and ready for you to take a look at.

templates/puppetdb-DLO-cleanup.service.epp
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Cleanup old discarded puppetdb reports
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/find /opt/puppetlabs/server/data/puppetdb/stockpile/discard/ -type f -mtime +90 -delete
```
austb: Also, one final note. For security reasons, this service should run as the [`PuppetDB user`](https://github.com/puppetlabs/puppetlabs-puppetdb#database_username)

Gemfile
bastelfreak: I rebased against the release branch to get all this fixed. I hope that was okay.

manifests/server.pp
bastelfreak: After months of work, this feature is finally released. It wasn't available when when I created this PR. updated the code.

manifests/server.pp
bastelfreak: sounds good, done

manifests/server.pp
bastelfreak: It needs to be enabled. Otherwise the timer is off after the next reboot and the upcoming puppet run would start it.

templates/puppetdb-DLO-cleanup.service.epp
bastelfreak: added user and group

bastelfreak:
Personal reminder:
* should add tests
* Update readme
* expose maximum DLO age as variable

austb:
Awesome! This looks good. If you need any help writing tests just let us know and we can help out. I would say these would be nice.

1. Can be disabled
1. Creates systemd service/timer, and starts timer
1. Creates cron job if systemd not available
1. Actually runs as puppetdb user. ie cannot perform some action the PuppetDB user should not be able to do (we can chat about this if it's unclear what to do, I'll have to think about it as well)

And if you add DLO age as variable, a test that that can be customized.

.travis.yml
```diff
@@ -1,31 +1,44 @@
...
 cache: bundler
-script: "bundle exec rake release_checks"
-#Inserting below due to the following issue: https://github.com/travis-ci/travis-ci/issues/3531#issuecomment-88311203
 before_install:
-  - gem update bundler
+  - bundle -v
+  - rm -f Gemfile.lock
+  - gem update --system
+  - gem --version
+  - bundle -v
```
rbrw: I'd suggest either moving these to a helper script,

```yaml
before_install: some/where/prep-tests
```

 or using a different approach, i.e. say with one of the yaml multi-line sytaxes.

For example, using ">":

```yaml
before_install: >
    bundle -v 
    && rm -f Gemfile.lock
    && ...
```
or using "|"

```yaml
before_install: |
    set -e
    bundle -v 
    rm -f Gemfile.lock
    ...
```
(Note that syntax is just approximate -- haven't checked it.)

Otherwise, travis will just keep going iirc.  That is, it runs all the bullet points, even if one of them fails, only stopping if there was a failure after it has tried them all.

.travis.yml
```diff
@@ -1,31 +1,44 @@
...
+script:
+  - 'bundle exec rake $CHECK'
```
rbrw: Hmm, do we need the single quotes, and regardless, wonder if we should quote the CHECK expansion as `"$CHECK"`, just to avoid having to worry about the content.

Zak-Kent:
@bastelfreak, sorry but would you be able to rebase this on master again? We ended up making some changes/rebasing the existing pr we had up before it merged. Thanks for all the help!

manifests/server.pp
bastelfreak: done

bastelfreak:
@austb can you review this again please?

austb:
Thanks @bastelfreak for the PR!

PR #280, opened 2018-07-14
use of hard-coded passwords is a bad practice 

akondasif:
> hard-coded default passwords as parameters is a bad practice

Greetings,

I am a security researcher, who is looking for security smells in Puppet scripts.
I noticed instances of hard-coded passwords, which are against the best practices
recommended by Common Weakness Enumeration (CWE) [https://cwe.mitre.org/data/definitions/259.html] and also by other security practitioners.
I suggest use of undef to mitigate this smell. Feedback is welcome.

### DIFF
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -1,0 +1,6 @@
+:hierarchy:
+  - common
+:backends:
+  - yaml
+:yaml:
+:datadir: 'hieradata'
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -1,0 +1,3 @@
+---
+db_pwd: puppetdb
+read_db_pwd: puppetdb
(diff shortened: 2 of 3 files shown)


austb:
Support for customizing your password via Hiera data already exists via use of a [control-repo](https://puppet.com/docs/pe/2018.1/control_repo.html). The [`hiera`](https://puppet.com/docs/puppet/5.2/hiera_use_hiera_functions.html) function is deprecated and will be removed shortly. The [`lookup`](https://puppet.com/docs/puppet/5.2/hiera_use_function.html) function is the preferred method for adding Hiera data to Puppet Code, but better options exists such as having it happen [automatically](https://puppet.com/docs/puppet/5.2/hiera_automatic.html). Adding Hiera data to this module won't happen, and defaulting to `undef` will create an authentication issue with PostgreSQL, see [`createuser`](https://www.postgresql.org/docs/9.1/static/app-createuser.html).

PR #285, opened 2018-09-18
(PDB-4092) Use vardir prefix for DLO path

austb:
Previously, the DLO path was hardcoded in with its default value. This
means that when attempting to clean a DLO store on a system using a
non-default location for vardir, the puppetdb module would attempt to
remove a directory that was not actually the DLO store (if it existed at
all).

Now, in accordance with how puppetdb selects the DLO store location, the
puppetdb module will use `<vardir>/stockpile/discard/` as the DLO
location.


Zak-Kent:
LGTM, thanks

PR #288, opened 2018-09-30
extend unit tests for main class

bastelfreak:


bastelfreak:
The error is:
```
  8) puppetdb on ubuntu-18.04-x86_64 with real world params should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile.with_all_deps }
       dependency cycles found: (Anchor[apt_key B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8 present] => Apt::Key[Add key: B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8 from Apt::Source apt.postgresql.org] => Apt::Source[apt.postgresql.org] => Package[dirmngr] => File[sources.list.d] => File[/etc/apt/sources.list.d/apt.postgresql.org.list] => Apt::Setting[list-apt.postgresql.org] => Apt::Source[apt.postgresql.org])
     # ./spec/unit/classes/init_spec.rb:37:in `block (5 levels) in <top (required)>'
```
This is a regression. I use the `puppetdb` class since months with these parameters, it now fails with the latest release/master branch.

bastelfreak:
I extended the unit tests a bit. The parameter that causes issues is `puppetdb::manage_dbserver` which calls out to the `postgresql` modules which uses the `apt` module.

bastelfreak:
The regression here is caused by https://github.com/puppetlabs/puppetlabs-apt/pull/801/files#diff-60ae41fd0a31977447947f59940ee9a4R215

eimlav:
Hey @bastelfreak I've reverted the PR with these changes. I hope this helps you with the issues you are encountering here!

bastelfreak:
Hi @eimlav, I updated and rebased this PR.The apt module is now working again. I think merging this PR still might be a good idea, to increase the code coverage.

austb:
Since `manage_dbserver` defaults to true, we were wondering if the shortened test would cover the case? I'm going to try to verify this locally, but you may be able to get it going faster.

```
next unless facts[:os]['release']['family'] == 'Debian'
context "on #{os} with managed postgresql" do
  describe 'with real world params' do
    it { is_expected.to compile.with_all_deps }
  end
  
  let :params do
    {
      :manage_dbserver => false
    }
  end

  describe 'with real world params' do
    it { is_expected.to compile.with_all_deps }
  end
end
```

bastelfreak:
That should be fine as well. I just wanted to use all params here that I use in prod and that currently fail for me. Afterwards I narrowed it down to `manage_dbserver`. On the other hand it might be good to test actual usecases in acceptance tests and not very specific parameters?

austb:
We were also cosidering removing the `next unless facts[:os]['release']['family'] == 'Debian'` so that if another dep breaks in postgresql, etc we can catch it as well.

austb:
I'm going to close this in favor of #289

PR #290, opened 2019-05-16
Allow inifile 3.x and postgresql 7.x

ekohl:


austb:
The travis failure looks like it just needs a bump in the ruby version

PR #295, opened 2019-06-14
allow newer versions of dependencies

mmoll:


mmoll:
a forge relase with this would be super nice...

austb:
Version 7.3.0 is available on the forge with this change https://forge.puppet.com/puppetlabs/puppetdb. Thanks again for this PR!

PR #296, opened 2019-07-24
Add possibilty to skip database creation in the module

jhooyberghs:
I want to be able to skip database creation in the module, so I have more flexibility to handle it in my own profile.


austb:
@jhooyberghs thanks for the contribution!

The outstanding work for this PR is adding tests for the new param and adding documentation to the README for the new parameter. Let me know if you'd like me to help with either of those.

austb:
@jhooyberghs this feature has been released in 7.4.0

PR #297, opened 2019-08-12
add possibility to set ip_mask_allow_all_users

aschaber1:
in case you need something else than 0.0.0.0/0,
for example ::/0.

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -1,15 +1,16 @@
 # Class for creating the PuppetDB postgresql database. See README.md for more
 # information.
 class puppetdb::database::postgresql(
-  $listen_addresses     = $puppetdb::params::database_host,
-  $database_name        = $puppetdb::params::database_name,
-  $database_username    = $puppetdb::params::database_username,
-  $database_password    = $puppetdb::params::database_password,
-  $database_port        = $puppetdb::params::database_port,
-  $manage_database      = $puppetdb::params::manage_database,
-  $manage_server        = $puppetdb::params::manage_dbserver,
-  $manage_package_repo  = $puppetdb::params::manage_pg_repo,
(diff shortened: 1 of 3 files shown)


manifests/database/postgresql.pp
```diff
@@ -1,15 +1,16 @@
...
-  $postgres_version     = $puppetdb::params::postgres_version,
+  $listen_addresses       = $puppetdb::params::database_host,
+  $database_name          = $puppetdb::params::database_name,
+  $database_username      = $puppetdb::params::database_username,
+  $database_password      = $puppetdb::params::database_password,
+  $database_port          = $puppetdb::params::database_port,
+  $manage_database        = $puppetdb::params::manage_database,
+  $manage_server          = $puppetdb::params::manage_dbserver,
+  $manage_package_repo    = $puppetdb::params::manage_pg_repo,
+  $postgres_version       = $puppetdb::params::postgres_version,
```
austb: Looks like we need one more space on these `=` above, so they line up with the one below.

manifests/server.pp
```diff
@@ -25,6 +25,7 @@
   $database_username                      = $puppetdb::params::database_username,
   $database_password                      = $puppetdb::params::database_password,
   $database_name                          = $puppetdb::params::database_name,
+  $database_ip_mask_allow_all_users       = $puppetdb::params::database_ip_mask_allow_all_users,
```
austb: The puppetdb module configures two things generally, the "database" (ie. the postgresql server and database) and the "server" which is the PuppetDB webserver. So this database configuration does not need to be in the `server.pp` parameters.

I think what we need instead is for this parameter to be added the `init.pp` and then passed into the class invocation of `puppetdb::database::postgresql`, which is at the end of `init.pp`.

PR #298, opened 2019-09-13
Create enable_storeconfigs option for puppet::master::config

gcoxmoz:
(MODULES-9905) Create enable_storeconfigs option for puppet::master::config

This creates a flag to allow the removal of 'storeconfigs' and 'storeconfigs_backend' directives from puppet.conf when those directives are managed by the module.

No tests are included here because there was no unit test to begin with(?!).

### DIFF
--- a/manifests/master/storeconfigs.pp
+++ b/manifests/master/storeconfigs.pp
@@ -3,7 +3,7 @@
 class puppetdb::master::storeconfigs (
   $puppet_conf = $puppetdb::params::puppet_conf,
   $masterless  = $puppetdb::params::masterless,
-  $enable      = false,
+  $enable      = true,
 ) inherits puppetdb::params {
 
   if $masterless {
--- a/manifests/master/storeconfigs.pp
+++ b/manifests/master/storeconfigs.pp
@@ -3,6 +3,7 @@
 class puppetdb::master::storeconfigs (
   $puppet_conf = $puppetdb::params::puppet_conf,
   $masterless  = $puppetdb::params::masterless,
+  $enable      = false,
 ) inherits puppetdb::params {
 
   if $masterless {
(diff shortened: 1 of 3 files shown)


austb:
Hi @gcoxmoz! Thanks for the contribution! Can you elaborate on your use case a bit? Generally I just want to make sure there's a reasonable time to be managing both PuppetDB and the Puppet Master's `storeconfigs` with the PuppetDB module, but _not_ sending the data to PuppetDB from Puppet runs.

gcoxmoz:
Sure, the use case here was to be able to temporarily sever contact between puppetserver and puppetdb during a prolonged outage of puppetdb.  Dropping 'puppetdb' out the 'reports' line was easy, but taking out 'storeconfigs' meant having to touch the puppet.conf on the server by hand instead of being able to turn it off in hiera.

manifests/master/storeconfigs.pp
```diff
@@ -3,6 +3,7 @@
 class puppetdb::master::storeconfigs (
   $puppet_conf = $puppetdb::params::puppet_conf,
   $masterless  = $puppetdb::params::masterless,
+  $enable      = false,
```
austb: I think this should default to `true` in case anyone applies it outside the scope of of `puppetdb::master::config` otherwise this would break their storeconfigs setting.

gcoxmoz: My rationale was that 'false' is [the puppet default](https://puppet.com/docs/puppet/6.0/configuration.html#storeconfigs) for 'storeconfigs'.   (IMO) it makes sense to be false in `storeconfigs.pp`, which is more like a private class.. which is then overridden by a call from `puppetdb::master::config` which pops it to 'true' in order to preserve module behavior for people using the module already.  Very little surprises, except the one you called out: people who have gone rogue.  I'd say this comes down to "is puppetdb::master::storeconfigs a private class?"

For principle-of-no-surprises I can see doing it your way, and this is not a hill I will choose to die on.  Consider this my one token piece of resistance; if you think yours is the better approach, I'll adjust accordingly.

gcoxmoz: Changed as requested based on no further discussion.

PR #300, opened 2020-02-14
Support CentOS 8, OracleLinux 8 and Debian 10 & support pl/inifile 4.x

ekohl:

PR #306, opened 2020-05-25
(maint)(MODULES-10112) bumping inifile dependency

h-haaks:


h-haaks:
This is my first PR to a puppet module, and I'm not the reporter of MODULES-10112.
I'm not sure if I'm supposed to update the jira issue.

austb:
Thank @h-haaks, I just merged this change in #300

PR #308, opened 2020-06-04
(maint) Update Puppet VS Code Extension ID

jpogran:
This commit updates the configuration file to point to the official Puppet VS Code Extension `puppet.puppet-vscode`


jpogran:
This change only affects users who open this repository inside VS Code and will not affect any module code. This is safe to merge. It is required to move users to the official Puppet VS Code extension

PR #310, opened 2020-08-02
Add option to set source ssl certs

kobybr:


austb:
@kobybr I trying to understand the motivation behind this change, it looks to do what the `puppetdb ssl-setup` script does. Is something about that script not working for you?

edestecd:
@austb I am not familiar with that script.  When is that script run? Is it done automatically upon install or when the service is started? 

Is it an issue to also have puppet manage the files in case they are updated etc?

PR #317, opened 2020-12-02
(maint) Test against puppet7 in travis

joshcooper:


Gemfile
```diff
@@ -36,6 +36,8 @@ group :system_tests do
   gem "serverspec",                                   require: false, platforms: [:ruby]
 end
 
+gem "scanf", require: false if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.7.0')
+
```
joshcooper, line 40: I'm thinking this is going to affect modules other than puppetdb, and would be better to add this to https://github.com/puppetlabs/puppet-module-gems/blob/117040802f7981aa097b16c03ffbc455d1f83e8f/config/dependencies.yml#L143-L155. I'll ping the modules folks

joshcooper: I think this is actually a puppet bug. The puppet gem should express a runtime gem dependency on `scanf`. Filed as https://tickets.puppetlabs.com/browse/PUP-10797

joshcooper: This will be released in puppet 7.1.0

joshcooper:
This is blocked on puppet 7.1.0 gem release

PR #322, opened 2021-03-03
(PDB-4764) Agent SSL certificates are used for communication with PostgreSQL

Filipovici-Andrei:
Updated the manifests for SSL configuration of PostgreSQL database configuration file.


manifests/params.pp
```diff
@@ -180,6 +180,14 @@
   $cleanup_timer_interval   = "*-*-* ${fqdn_rand(24)}:${fqdn_rand(60)}:00"
   $dlo_max_age              = 90
 
+  # certificats used for PostgreSQL SSL configuration. Puppet certificates are used
+  $postgresql_ssl_on            = true
+  $postgresql_ssl_folder        = "${puppet_confdir}/ssl"
```
austb, line 185: After removing the `+=` it fails here because the postgresql user doesn't have permission to read the agent certificates. I think we'll need to copy them to the postgresql user.

manifests/server/database.pp
```diff
@@ -87,6 +88,14 @@
 
     $subname = "//${database_host}:${database_port}/${database_name}${database_suffix}"
 
+    if $postgresql_ssl_on {
+      $subname += @("EOT")
```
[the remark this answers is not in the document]
austb: I also noticed that this was creating multiple lines in the config file, so we might just have to deal with a really long string unless there's another way to do it that creates a single line.

manifests/server/database.pp
```diff
@@ -85,7 +89,16 @@
       $database_suffix = ''
     }
 
-    $subname = "//${database_host}:${database_port}/${database_name}${database_suffix}"
+    $subname_default = "//${database_host}:${database_port}/${database_name}${database_suffix}"
+
+    if $postgresql_ssl_on {
+      $subname = "${subname_default}?\
```
austb, line 95: This seems to be creating large spaces in the conf file, after the first space it seems to stop reading any data.

manifests/server/database.pp
```diff
@@ -85,7 +89,16 @@
       $database_suffix = ''
     }
 
-    $subname = "//${database_host}:${database_port}/${database_name}${database_suffix}"
+    $subname_default = "//${database_host}:${database_port}/${database_name}${database_suffix}"
```
austb, line 92: The default `database_host` is `localhost`, but for the certificate to validate I needed to make it be the `fqdn`.

One tricky thing is that the default binding of the postgresql service is to 127.0.0.1, but for the FQDN to resolve it needs to be bound to `0.0.0.0`. As a hack while I was testing I added an entry to `/etc/hosts` but I'm not sure that's a viable solution for the module.

manifests/database/postgresql.pp
```diff
@@ -24,6 +28,62 @@
...
+        require => [File['postgres private key'], Concat['postgres cert bundle']]
+      }
+
+      postgresql::server::config_entry {'ssl_cert_file':
+        ensure  => present,
+        value   => "${postgresql::params::datadir}/server.crt",
+        require => [File['postgres private key'], Concat['postgres cert bundle']]
+      }
+
+      # postgresql::server::config_entry {'ssl_ca_file':
```
austb, line 80: I eventually got to the state where postgresql required this setting with the error message
```
2021-03-08 22:23:15.457 UTC [17580] LOG:  client certificates can only be checked if a root certificate store is available
2021-03-08 22:23:15.457 UTC [17580] HINT:  Make sure the configuration parameter "ssl_ca_file" is set.
2021-03-08 22:23:15.457 UTC [17580] CONTEXT:  line 31 of configuration file "/etc/postgresql/9.6/main/pg_hba.conf"
2021-03-08 22:23:15.457 UTC [17580] LOG:  client certificates can only be checked if a root certificate store is available
2021-03-08 22:23:15.457 UTC [17580] HINT:  Make sure the configuration parameter "ssl_ca_file" is set.
2021-03-08 22:23:15.457 UTC [17580] CONTEXT:  line 36 of configuration file "/etc/postgresql/9.6/main/pg_hba.conf"
2021-03-08 22:23:15.457 UTC [17580] FATAL:  could not load pg_hba.conf
```

austb (commented):
To document the settings I had to add for client authentication I wanted to add these here

`pg_hba.conf`
```
# Rule Name: Allow certificate mapped connections to pe-puppetdb as pe-puppetdb (ipv4)
# Description: none
# Order: 0
hostssl puppetdb     puppetdb     0.0.0.0/0       cert    map=puppetdb-puppetdb-map clientcert=1

# Rule Name: Allow certificate mapped connections to pe-puppetdb as pe-puppetdb (ipv6)
# Description: none
# Order: 1
hostssl puppetdb     puppetdb     ::/0    cert    map=puppetdb-puppetdb-map clientcert=1
```

This is a file I forgot about, buy `pg_ident.conf` defines the certificate name to database username mappings
```
# This file is managed by Puppet. DO NOT EDIT.
puppetdb-puppetdb-map duller-artery.delivery.puppetlabs.net puppetdb
```

A lot of my configuration to get it working was done manually, so in case I missed anything I've extended the lifetime of `duller-artery.delivery.puppetlabs.net` in case you want to dig around in the settings.

austb (commented):
This looks good, I think it could use a couple of basic tests for the important settings, like pg_ident/pg_hba rules, the subname config setting, and the pk8 and then it should be ready to go.

PR #327, opened 2021-04-09
allow current versions of inifile, firewall, stdlib

kenyon:
`metadata.json` update to allow the current versions of the inifile, firewall, and stdlib modules.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/inifile",
-      "version_requirement": ">= 1.1.3 < 5.0.0"
+      "version_requirement": ">= 1.1.3 < 6.0.0"
     },
     {
       "name": "puppetlabs/postgresql",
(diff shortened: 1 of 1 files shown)


smortex:
@kenyon stumble here after following links from other issue… I guess you need to sigh the CLA above for this issue to be fixed?

smortex (approved):

kenyon:
Still waiting on my company to allow me to sign the CLA. I think this one can fall under the trivial patch exemption though.

ekohl (approved):
I would think so too. There's literally no other way to do this so this feels trivial. However, not a lawyer and I don't work for Puppet.

austb:
Yeah, sorry I was off for a little while. This PR falls under the trivial patch exemption. I'll merge it, and try to get a release out in the next few days. There are some issues with the module's CI that I'll need to resolve before the release can go out.

austb (approved):

PR #330, opened 2021-06-09
(maint) Add read-only user.

Filipovici-Andrei:
There used to be only one user (puppetdb) which was used for all operations on the database.
This PR adds a read only user in PostgreSQL which will be used only for queries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -608,15 +608,15 @@ Which database backend to use for the read database. Only supports
 #### `read_database_host`
 *This parameter must be set to use another PuppetDB instance for queries.*
 
-The hostname or IP address of the read database server. If set to `undef`, 
-it will use the value of the `database_host` parameter. This option is
-supported in PuppetDB >= 1.6.
+The hostname or IP address of the read database server. If set to `undef`, and 
+`manage_database` is set to `true`, it will use the value of the `database_host` 
+parameter. This option is supported in PuppetDB >= 1.6.
 
 #### `read_database_port`
 
 The port that the read database server listens on. If `read_database_host`
-is set to `undef`, it will use the value of the `database_port` parameter. This
-option is supported in PuppetDB >= 1.6.
+is set to `undef`, and `manage_database` is set to `true`, it will use the value of 
+the `database_port` parameter. This option is supported in PuppetDB >= 1.6.
 
 #### `read_database_username`
 
@@ -637,8 +637,8 @@ Defaults to `true`
 #### `read_database_name`
 
 The name of the read database instance to connect to. If `read_database_host`
-is set to `undef`, it will use the value of the `database_name` parameter.
-This option is supported in PuppetDB >= 1.6.
+is set to `undef`, and `manage_database` is set to `true`, it will use the value of
+the `database_name` parameter. This option is supported in PuppetDB >= 1.6.
 
 #### `read_log_slow_statements`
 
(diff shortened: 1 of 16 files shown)


Filipovici-Andrei:
Because the read only user is added in the public role, it is able to create and delete it's own tables

README.md
```diff
@@ -606,25 +606,26 @@ Which database backend to use for the read database. Only supports
...
 
-The port that the read database server listens on. Defaults to `5432`. This
+The port that the read database server listens on. If `read_database_host`
+is set to `undef`, it will use the value of the `database_port` parameter. This
 option is supported in PuppetDB >= 1.6.
 
 #### `read_database_username`
 
-The name of the read database user to connect as. Defaults to `puppetdb`. This
+The name of the read database user to connect as. Defaults to `puppetdb-read`. This
```
IrimieBogdan, line 623: Does this always default to `puppetdb-read`. If `managed` is set to false will it still default to `puppetdb-read` or will it default to `database_username`?

README.md
```diff
@@ -606,25 +606,26 @@ Which database backend to use for the read database. Only supports
...
 option is supported in PuppetDB >= 1.6.
 
 #### `read_database_password`
 
-The password for the read database user. Defaults to `puppetdb`. This option is
+The password for the read database user. Defaults to `puppetdb-read`. This option is
```
IrimieBogdan, line 628: Same as for username

IrimieBogdan (approved):

README.md
Filipovici-Andrei: This always defaults to puppetdb-read

README.md
Filipovici-Andrei: Yes, this will also default to puppetdb-read as well.

spec/unit/classes/database/ssl_configuration_spec.rb
```diff
@@ -135,5 +149,44 @@
           .with_database_username(params[:database_name])
       end
     end
+
+    context 'when the create_read_user_rule is set to true' do
+      let(:params) do
+        {
+          database_name: 'puppetdb',
+          read_database_username: 'puppetdb-read',
+          create_read_user_rule: true,
```
austb, line 158: I added the comma here to appease rubocop.

austb (approved):

alexjfisher:
FWIW, I think this should have been labelled as a breaking change, not just a feature. If you weren't managing postgresql with this module, the change in puppetdb configuration stops puppetdb from starting (my case).
If you were, then your database now gets a user with a non-strong default password?

waipeng:
I wonder if anyone is running into the following error?

Error: /Stage[main]/Puppetdb::Database::Postgresql/Puppetdb::Database::Read_only_user[puppetdb-read]/Puppetdb::Database::Default_read_grant[puppetdb grant read permission on new objects from puppetdb to puppetdb-read]/Postgresql_psql[grant default select permission for puppetdb-read]: Could not evaluate: Error evaluating 'unless' clause, returned pid 22961 exit 2: 'sh: 8: Syntax error: Unterminated quoted string  

Looks like there are mismatch quotes in the code.

I've also opened up a ticket in puppet https://tickets.puppetlabs.com/browse/MODULES-11204, as "Issues" is not enabled for this repo in Github.

PR #331, opened 2021-06-29
Allow `Sensitive` data type for secrets

cocker-cc:


manifests/database/postgresql.pp
```diff
@@ -70,10 +70,22 @@
   }
 
   if $manage_database {
+    # TODO: Workaround, until "postgresql" accepts Sensitive
```
smortex, line 73: It has been released in puppetlabs-postgresql 7.5.0 thanks to your work on the topic, thanks!
https://github.com/puppetlabs/puppetlabs-postgresql/pull/1279

This will need a bump of the minimal module version in `metadata.conf`.

cocker-cc: I removed the Workarounds, and adapted the Versions in `metadata.json`.

metadata.json
```diff
@@ -10,11 +10,11 @@
   "dependencies": [
     {
       "name": "puppetlabs/inifile",
-      "version_requirement": ">= 1.1.3 < 6.0.0"
+      "version_requirement": ">= 5.1.0 < 6.0.0"
```
smortex, line 13: This looks unrelated... Is this one required?

smortex: Ha! Excellent, thanks!

smortex (approved):
LGTM!

manifests/database/postgresql.pp
```diff
@@ -16,7 +16,7 @@
   $postgresql_ssl_cert_path    = $puppetdb::params::postgresql_ssl_cert_path,
   $postgresql_ssl_ca_cert_path = $puppetdb::params::postgresql_ssl_ca_cert_path,
   $read_database_username      = $puppetdb::params::read_database_username,
-  $read_database_password      = $puppetdb::params::read_database_password,
+  Optional[Variant[String, Sensitive[String]]] $read_database_password = $puppetdb::params::read_database_password,
```
bastelfreak, line 19: I think `Optional[]` doesn't make sense here. The module has a default, `$puppetdb::params::read_database_password`, with the value `puppetdb-read`. You cannot overwrite it with undef.

cocker-cc: You are right.  I fixed and force-pushed it.  
86af360e6b496c4d75b5491e51ef55f2ba378554

manifests/init.pp
```diff
@@ -62,7 +62,7 @@
   $read_database_host                      = $puppetdb::params::read_database_host,
   $read_database_port                      = $puppetdb::params::read_database_port,
   $read_database_username                  = $puppetdb::params::read_database_username,
-  $read_database_password                  = $puppetdb::params::read_database_password,
+  Optional[Variant[String, Sensitive[String]]] $read_database_password = $puppetdb::params::read_database_password,
```
smortex, line 65:
```suggestion
  Variant[String, Sensitive[String]] $read_database_password = $puppetdb::params::read_database_password,
```

manifests/server.pp
```diff
@@ -51,7 +51,7 @@
   $read_database_host                      = $puppetdb::params::read_database_host,
   $read_database_port                      = $puppetdb::params::read_database_port,
   $read_database_username                  = $puppetdb::params::read_database_username,
-  $read_database_password                  = $puppetdb::params::read_database_password,
+  Optional[Variant[String, Sensitive[String]]] $read_database_password = $puppetdb::params::read_database_password,
```
smortex, line 54:
```suggestion
  Variant[String, Sensitive[String]] $read_database_password = $puppetdb::params::read_database_password,
```

manifests/server/read_database.pp
```diff
@@ -4,7 +4,7 @@
   $read_database_host     = $puppetdb::params::read_database_host,
   $read_database_port     = $puppetdb::params::read_database_port,
   $read_database_username = $puppetdb::params::read_database_username,
-  $read_database_password = $puppetdb::params::read_database_password,
+  Optional[Variant[String, Sensitive[String]]] $read_database_password = $puppetdb::params::read_database_password,
```
smortex, line 7:
```suggestion
  Variant[String, Sensitive[String]] $read_database_password = $puppetdb::params::read_database_password,
```

smortex (commented):
A few more extra `Optional[...]` seems to be present.

Also, while we are improving this, what about making the default Sensitive too in `params.pp`?

`$foo = Sensitive('bar')`

cocker-cc:
> A few more extra `Optional[...]` seems to be present.

I removed `Optional`.  Indeed it was unnecessary.

> Also, while we are improving this, what about making the default Sensitive too in `params.pp`?
> `$foo = Sensitive('bar')`

I do not see this necessary, as the Default-Value is publicly visible anyway.

smortex:
> I do not see this necessary, as the Default-Value is publicly visible anyway.

My understanding is that it helps to have Puppet automatically redact secrets in diff, and that at some point only a Sensitive would be accepted in a future major version.  That being said, we have default passwords in this module which is not a best practice so maybe this will not happen before a loooong time :smile:.

I am fine with the PR as it is, so will let other reviewers tell what they think!

Thank you!

manifests/database/postgresql.pp
```diff
@@ -75,7 +75,7 @@
   $puppetdb_server             = $puppetdb::params::puppetdb_server,
   $database_name               = $puppetdb::params::database_name,
   $database_username           = $puppetdb::params::database_username,
-  $database_password           = $puppetdb::params::database_password,
+  Variant[String, Sensitive[String]] $database_password = $puppetdb::params::database_password,
```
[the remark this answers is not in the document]
cocker-cc: done

bastelfreak (approved):

h0tw1r3 (approved):

PR #333, opened 2021-07-18
(PDB-5052) Install PostgreSQL 11 for PDB > 7.0.0

oanatmaria:


manifests/params.pp
```diff
@@ -24,7 +24,12 @@
   } else {
     $manage_pg_repo            = false
   }
-  $postgres_version          = '9.6'
+
+  if $puppetdb_version in ['latest','present'] or versioncmp($puppetdb_version, '7.0.0') >= 0 {
+    $postgres_version          = '11.12'
```
austb, line 29: Between postgres 9.6 and postgres 10 they changed their versioning semantics. Before the third number indicated a minor release, whereas now incrementing the second number is a minor release. I think we can just use `11` here to ensure that we pick up the latest minor releases of postgres as they are release.

More details of their versioning policy can be found here https://www.postgresql.org/support/versioning/

austb (approved):

PR #337, opened 2021-10-05
(SERVER-2500) Use DNF modules to install postgres on RHEL 8

Magisus:
The postgresql module added support for DNF modules to install postgres
at specific versions on RHEL 8. However in order to install a
non-default version, like we need to test Puppet Platform 6, which
only supports Postgres 9.6, `manage_dnf_module` must be set to true for
the `puppetlabs-postgresql` module. This commit ensures that mode is set
when isntalling on RHEL 8.

Note that when the default version is being installed (postgres 10 on
RHEL8), the code path that this flag opts into is already being used. So
this should only affect Platform 6 users on RHEL 8, specifically.


Magisus:
For the postgresql module change, see https://github.com/puppetlabs/puppetlabs-postgresql/pull/1239.

This resolves the follow issue seen when trying to enable install PDB with this module on RHEL 8 in our acceptance tests:
```
Error: Execution of '/usr/bin/dnf -d 0 -e 1 -y install postgresql96-server' returned 1: Error: Unable to find a match
```
This is because the platform string is invalid for `dnf` version selection, we need to be using the top branch in [this conditional](https://github.com/puppetlabs/puppetlabs-postgresql/blob/8e040de1488c1cada00763e550a59284da15d2e4/manifests/params.pp#L44)

Magisus:
I wasn't sure if this should be exposed to users or not.

Magisus:
I tested this on 7.x and was surprised to see us still trying to install postgres 9.6 there... then I realized this module hasn't been released in a while, so we were not getting https://github.com/puppetlabs/puppetlabs-puppetdb/commit/3ea57a587d33e3105962d0e5f4f4b3196feb48c3. I think this is necessary even after that goes in, though, since we're still using a non-default version (11 instead of 10).

bschonec:
I'm returning to the world of Puppet after a four year break.  Is there any reason this project is tied to Postgresql 9.6 and 11?  I've installed PSQL 10, 11 and 12 and (so far) haven't had any issues.

jhoblitt:
psql 11 is EOL and 12 is the oldest version that we need to worry about supporting.

manifests/params.pp
```diff
@@ -25,6 +25,12 @@
     $manage_pg_repo            = false
   }
 
+  if $::osfamily =~ /RedHat/ and $facts['os']['release']['major'] == '8' {
```
jhoblitt, line 28: This should be changed to check for either any version >= 8 or match EL8 & EL9.

jhoblitt (changes requested):

PR #339, opened 2021-11-28
fix unless clause when grant default permissions

igomura:
Please check out the suggested PR. It fixes the case when we need to have RO user for the database.

### DIFF
--- a/manifests/database/default_read_grant.pp
+++ b/manifests/database/default_read_grant.pp
@@ -19,7 +19,7 @@ define puppetdb::database::default_read_grant(
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*(\\\\\")?${database_read_only_username}(\\\\\")?=r/(\\\\\")${database_username}(\\\\\")?.*'
+                WHERE acl.defaclacl::text ~ '.*(\\\\\")?${database_read_only_username}(\\\\\")?=r/(\\\\\")?${database_username}(\\\\\")?.*'
                 AND nspname = '${schema}'",
   }
 
--- a/manifests/database/default_read_grant.pp
+++ b/manifests/database/default_read_grant.pp
@@ -19,7 +19,7 @@ define puppetdb::database::default_read_grant(
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=r/${database_username}\\\".*'
+                WHERE acl.defaclacl::text ~ '.*(\\\\\")?${database_read_only_username}(\\\\\")?=r/(\\\\\")${database_username}(\\\\\")?.*'
                 AND nspname = '${schema}'",
   }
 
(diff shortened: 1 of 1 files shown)


igomura:
@Filipovici-Andrei please review

igomura:
@thde please hold on. I've realized my PR is for the corner case too. The regex expects doubled quotes around the usernames but it is not always the case. It only happens if username has dashes or any other non-alphabetical symbols. Without them there will be no quotes. So, I have to rework my regex to match both cases

igomura:
now it works for both cases. the PR is ready for review

manifests/database/default_read_grant.pp
```diff
@@ -19,7 +19,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=r/${database_username}\\\".*'
+                WHERE acl.defaclacl::text ~ '.*(\\\\\")?${database_read_only_username}(\\\\\")?=r/(\\\\\")${database_username}(\\\\\")?.*'
```
smortex, line 22: Did you mean:

```suggestion
                WHERE acl.defaclacl::text ~ '.*(\\\\\")?${database_read_only_username}(\\\\\")?=r/(\\\\\")?${database_username}(\\\\\")?.*'
```

In seems to be missing only on this query.

smortex: Also, I think you are expecting balanced quotes, one of:
1. `foo=r/bar`
2. `foo=r/"b-a-r"`
3. `"f-o-o"=r/bar`
4. `"f-o-o"=r/"b-a-r"`

In this case you can use back-reference to previous captures:
```suggestion
                WHERE acl.defaclacl::text ~ '.*((?:\\\\\")?)${database_read_only_username}\\1=r/((?:\\\\\")?)${database_username}\\2.*'
```

The leading and trailing `.*` can also be removed, but it might make more sense to match the full field:

```suggestion
                WHERE acl.defaclacl::text ~ '^{"((?:\\\\\")?)${database_read_only_username}\\1=r/((?:\\\\\")?)${database_username}\\2"}$'
```

The same apply for the other 2 queries.

igomura: Thank you @smortex for corrections. I'm totally agreed with the first one but two others I have to test. I'm not a great expert in regex.

mookie-: For me, there were some additional `\` needed due to the escaping hell from puppet to bash to psql.

This worked for me:
```
                WHERE acl.defaclacl::text ~ '^{\"((?:\\\\\\\\\")?)${database_read_only_username}\\1=r\\/((?:\\\\\\\\\")?)${database_username}\\2\"}$'
                WHERE acl.defaclacl::text ~ '^{\"((?:\\\\\\\\\")?)${database_read_only_username}\\1=U\\/((?:\\\\\\\\\")?)${database_username}\\2\"}$'
                WHERE acl.defaclacl::text ~ '^{\"((?:\\\\\\\\\")?)${database_read_only_username}\\1=X\\/((?:\\\\\\\\\")?)${database_username}\\2\"}$'
```

elkangaroo:
I am getting the error as mentioned in https://github.com/puppetlabs/puppetlabs-puppetdb/pull/330#issuecomment-935496488 when using 
```
class { 'puppetdb':
        manage_dbserver        => false
```

Is there a workaround until this gets merged + released or do I need to downgrade the module to 7.8.0?

waipeng:
Hi all, may I check what's happening to this? Is it just waiting for @igomura to sign the CLA?

It seems a waste that all these effort went into coming up with a PR and reviews but it got stuck here. Any suggestions on how we can move this forward?

Alternatively, does anyone have a workaround?

smortex:
Closing this one as @waipeng opened #371 to continue working on this using @jhunt-steds suggestion which improve readability.

PR #344, opened 2022-02-09
Update the module using the PDK

smortex:
Update with a recent PDK with the latest templates.

Allows up to have working CI.


spec/unit/classes/database/ssl_configuration_spec.rb
```diff
@@ -184,7 +187,7 @@
       it 'has read ident rule' do
         is_expected.to contain_postgresql__server__pg_ident_rule("Map the SSL certificate of the server as a #{params[:read_database_username]} user")
           .with_map_name(read_identity_map)
-          .with_system_username(facts[:fqdn])
+          .with_system_username(facts[:networking][:fqdn])
```
bastelfreak, line 190: shouldn't that be:
```suggestion
          .with_system_username(facts[:networking]['fqdn'])
```

smortex: Humm… it seems to be worse.  That being said at the beginning of the file we have:

```ruby
      let(:facts) do
        {
          # ...
          networking: {
            fqdn: 'cheery-rime@puppet',
          },
          # ...
        }
      end
```

This module is hell… :scream:

smortex:
Current state: out of 36 failing unit tests, 12 are due to the module saying it support Ubuntu 16.04 which is EOL and for which the postgresql module drop support more than 1 year ago.  I'll try to fix the other unit tests, when done will in another PR to drop EOL OS.

smortex:
LGTM regarding unit tests.

I will drop Ubuntu 16.04 which has been EOL for 2.5 years in another PR and the unit tests should be green.

Because the module has been lagging behind for years, I'll wait for this to be merged before investing more time in fixing the tests.

bastelfreak (approved):

h0tw1r3:
all changes were cherry-picked into #385

PR #350, opened 2022-02-23
postgresql version >=10 onwards does not require contrib package

KoenDierckx:
Fix for https://tickets.puppetlabs.com/browse/PDB-5454

_Started a new PR (to avoid license issue)_

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -59,8 +59,9 @@ class puppetdb::database::postgresql (
     # Only install pg_trgm extension, if database it is actually managed by the module
     if $manage_database {
 
-      # from postgresql version 10 onwards, this extension is no longer inside the contrib package, but is being bundled with the postgresql package itself
-      if (versioncmp($postgres_version, '10') < 0) {
+      # from postgresql version 10 onwards, this extension is no longer inside
+      # the contrib package, but is being bundled with the postgresql package itself
+      if versioncmp($postgres_version, '10') < 0 {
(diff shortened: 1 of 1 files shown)


manifests/database/postgresql.pp
```diff
@@ -59,8 +59,11 @@
     # Only install pg_trgm extension, if database it is actually managed by the module
     if $manage_database {
 
-      # get the pg contrib to use pg_trgm extension
-      class { '::postgresql::server::contrib': }
+      # from postgresql version 10 onwards, this extension is no longer inside the contrib package, but is being bundled with the postgresql package itself
+      if (versioncmp($postgres_version, '10') < 0) {
```
kenyon, line 63:
```suggestion
      # from postgresql version 10 onwards, this extension is no longer inside
      # the contrib package, but is being bundled with the postgresql package itself
      if versioncmp($postgres_version, '10') < 0 {
```

ekohl (changes requested):
I believe this PR is invalid: https://github.com/puppetlabs/puppetlabs-postgresql/commit/10b5d37f6944fa2c2d28f78fdbdfc6d8a3d0acd2 resolved this by making the contrib class a noop if there is no package. This keeps it compatible with distributions where there is still a contrib package.

ekohl:
Oh, I just realized that's in 8.0.0 and this module isn't marked as compatible so I opened https://github.com/puppetlabs/puppetlabs-puppetdb/pull/353.

PR #351, opened 2022-03-16
cater for multiple server_urls 

bassonj:
Documentation at [puppet.com](https://puppet.com/docs/puppetdb/7/puppetdb_connection.html#server_urls) states that multiple server_urls can be specified

> You can use a comma-separated list of URLs if there are multiple PuppetDB instances available. A server_urls config that supports two PuppetDBs would look like:

`server_urls = https://puppetdb1.example.com:8081,https://puppetdb2.example.com:8081`

This MR changes the `puppet_server` String into an Array and runs .map on it to allow for multiple puppetdb servers to be specified

### DIFF
--- a/lib/puppet/util/puppetdb_validator.rb
+++ b/lib/puppet/util/puppetdb_validator.rb
@@ -24,30 +24,28 @@ class Puppet::Util::PuppetdbValidator
     end
   end
 
-  def valid_connection_new_client?
-    puppetdb_server.each { | server |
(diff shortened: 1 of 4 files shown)


binford2k:
Also, we probably want a couple spec updates in https://github.com/puppetlabs/puppetlabs-puppetdb/tree/main/spec/unit/classes/master to validate behaviors

bassonj:
I've updated the approach to deal with either a String or an Array of strings so no regression is introduced.
I've also updated the specs to the best of my abilities.

Please let me know if there's anything else that needs to happen to expedite the acceptance of this change

PR #354, opened 2022-04-23
Revert "postgresql version >=10 onwards does not require contrib package"

bastelfreak:
Reverts puppetlabs/puppetlabs-puppetdb#350

reverting because it was merfed by accident


crazymind1337:
can we have a new release for that please soon? this bug caused an incident in our setup.

crazymind1337:
> can we have a new release for that please soon? this bug caused an incident in our setup.

Nevermind

PR #358, opened 2022-09-16
Set owner of server config.ini to root

ekohl:
The puppetdb user doesn't need to write to the config (it's managed by Puppet after all) so setting the owner to root and using the group to only read is safer.

It is also closer to packaging, which ensures the owner is root after a package update. Ideally packaging would use the same group and mode as well so there isn't an event after a package update, but this change at least trims it down from 3 changes to 2.

### DIFF
--- a/manifests/server/global.pp
+++ b/manifests/server/global.pp
@@ -2,7 +2,6 @@
 class puppetdb::server::global (
   $vardir         = $puppetdb::params::vardir,
   $confdir        = $puppetdb::params::confdir,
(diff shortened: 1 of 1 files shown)


ekohl:
> It is also closer to packaging, which ensures the owner is root after a package update. Ideally packaging would use the same group and mode as well so there isn't an event after a package update, but this change at least trims it down from 3 changes to 2.

Any idea where the RPM spec for this is so I can submit a PR?

bastelfreak (approved):

PR #359, opened 2022-10-17
set encoding when creating the DB

elfranne:
According to the [docs](https://puppet.com/docs/puppetdb/7/configure_postgres.html) you should specify the encoding when creating the database. The database encoding will be set depending on your locale (I got LATIN1) unless specified.

I am getting some errors is the PuppetDB logs: 
```
ERROR: unsupported Unicode escape sequence Detail: Unicode escape values cannot be used for code point values above 007F when the server encoding is not UTF8.
```
and filling up few gig of additional log at `/opt/puppetlabs/server/data/puppetdb/stockpile/discard`

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -149,6 +149,8 @@ class puppetdb::database::postgresql (
     postgresql::server::db { $database_name:
       user     => $database_username,
       password => $database_password,
+      encoding => 'UTF8',
+      locale   => 'en_US.UTF-8',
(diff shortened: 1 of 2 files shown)


elfranne:
This only applies on newly created puppetdb. If you don t care about the data in the db you can: 

```puppet
sudo service puppetdb stop
sudo -i -u postgres psql
DROP DATABASE puppetdb;

puppet apply manifest/site.pp
```

bastelfreak (approved):

PR #361, opened 2022-12-07
(PDB-5559) Grant read user to write user

austb:
To ensure old report partitions are removed promptly without blocking commands and queries the write user needs the ability to cancel running puppetdb queries. To do that the write user needs to be granted the read user role.


jpartlow (approved):
:+1: Tested out the sql manually, and the postgresql_sql resource looks correct.

PR #362, opened 2023-04-19
(PDB-5611) Update legacy facts to structured form

austb:


kenyon (approved):

jpartlow (commented):
We should probably clean up the facts in the specs as well. Since the specs aren't failing, maybe just deleting the legacy facts? Though I suppose if there are specs that pretend to care about osfamily or fqdn, they might need some prodding.

PR #363, opened 2023-04-26
fix incorrect year in release date

corporate-gadfly:


kenyon:
Changelog should be automatically generated, so not sure how this would become wrong.

corporate-gadfly:
Not that important, so closing.

smortex:
That was fixed in 0053513364d07f0c78ade801bcdcf322abfbab18.  It looks like this changelog is crafted by hand :scream:

corporate-gadfly:
Ahh... Makes sense, TY.

PR #370, opened 2023-08-30
Prefer `clientcert=verify-full` to the deprecated `clientcert=1` in hba rules

marszip:
We have updated to Puppet 8.2 and in the release notes it states that PostgreSQL 13 is not supported. The most current version for us is PostgreSQL 15, but we run into a problem with the Puppetforge module for Puppet DB.

The clientcert=1 parameter in the pg_hba.conf is not valid anymore.
This should be clientcert=verify-ca or clientcert=verify-full.

### DIFF
--- a/manifests/database/postgresql_ssl_rules.pp
+++ b/manifests/database/postgresql_ssl_rules.pp
@@ -13,7 +13,7 @@ define puppetdb::database::postgresql_ssl_rules (
     address     => '0.0.0.0/0',
     auth_method => 'cert',
     order       => 0,
-    auth_option => "map=${identity_map_key} clientcert=1"
+    auth_option => "map=${identity_map_key} clientcert=verify-full"
   }
 
   postgresql::server::pg_hba_rule { "Allow certificate mapped connections to ${database_name} as ${database_username} (ipv6)":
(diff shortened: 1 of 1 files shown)


smortex (approved):
`clientcert=verify-full` (or `verify-ca`) was introduced in PostgreSQL 12, so all supported versions of PostgreSQL support this setting.  It seems safe to update it.

h0tw1r3:
> `clientcert=verify-full` (or `verify-ca`) was introduced in PostgreSQL 12, so all supported versions of PostgreSQL support this setting. It seems safe to update it.

Unless I'm reading this wrong, Postgresql 11 is usually installed by default.
 
https://github.com/puppetlabs/puppetlabs-puppetdb/blob/0053513364d07f0c78ade801bcdcf322abfbab18/manifests/params.pp#L29

smortex:
> > `clientcert=verify-full` (or `verify-ca`) was introduced in PostgreSQL 12, so all supported versions of PostgreSQL support this setting. It seems safe to update it.
> 
> Unless I'm reading this wrong, Postgresql 11 is usually installed by default.
> 
> https://github.com/puppetlabs/puppetlabs-puppetdb/blob/0053513364d07f0c78ade801bcdcf322abfbab18/manifests/params.pp#L29

This should probably be changed.  These versions have reached EOL a long time ago.

As far as I am concerned, I set `manage_dbserver => false` in my control-repo's puppetdb profile.  Managing the server from this module is probably a smell.

h0tw1r3:
closing in favor of #380

PR #371, opened 2023-09-01
Fix: Unterminated quoted string when creating read user

waipeng:
puppetdb in default config will create a read-only user, however there is a bug[1] with the syntax to set the default read grant.

Fix it with help from comments[2].

[1] https://github.com/puppetlabs/puppetlabs-puppetdb/pull/330#issuecomment-935496488
[2] https://github.com/puppetlabs/puppetlabs-puppetdb/pull/339#issuecomment-1163552126

### DIFF
--- a/manifests/database/default_read_grant.pp
+++ b/manifests/database/default_read_grant.pp
@@ -20,7 +20,7 @@ define puppetdb::database::default_read_grant (
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=r/${database_username}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=r/(\"?)${database_username}\\2@'
                 AND nspname = '${schema}'",
   }
 
@@ -37,7 +37,7 @@ define puppetdb::database::default_read_grant (
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=U/${database_username}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=U/(\"?)${database_username}\\2@'
                 AND nspname = '${schema}'",
   }
 
@@ -54,7 +54,7 @@ define puppetdb::database::default_read_grant (
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=X/${database_username}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=X/(\"?)${database_username}\\2@'
                 AND nspname = '${schema}'",
   }
 }
--- a/spec/support/unit/shared/database.rb
+++ b/spec/support/unit/shared/database.rb
@@ -72,7 +72,7 @@ shared_examples 'postgresql_psql default read grant' do
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"#{with[:database_read_only_username]}\\\\\"=r/#{with[:database_username]}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)#{with[:database_read_only_username]}\\1=r/(\"?)#{with[:database_username]}\\2@'
                 AND nspname = 'public'",
       )
   }
@@ -92,7 +92,7 @@ shared_examples 'postgresql_psql default read grant' do
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"#{with[:database_read_only_username]}\\\\\"=U/#{with[:database_username]}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)#{with[:database_read_only_username]}\\1=U/(\"?)#{with[:database_username]}\\2@'
                 AND nspname = 'public'",
       )
   }
@@ -112,7 +112,7 @@ shared_examples 'postgresql_psql default read grant' do
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"#{with[:database_read_only_username]}\\\\\"=X/#{with[:database_username]}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)#{with[:database_read_only_username]}\\1=X/(\"?)#{with[:database_username]}\\2@'
                 AND nspname = 'public'",
       )
   }


waipeng:
There are no tests, but I can confirm this works.

I am checking out the situation with the tests.

manifests/database/default_read_grant.pp
```diff
@@ -19,7 +19,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=r/${database_username}\\\".*'
+                WHERE array_to_string(acl.defaclacl, ' + ') ~ '.*(\")?${database_read_only_username}(\")?=r/(\")?${database_username}(\")?.*'
```
smortex, line 22: The multiple layers of escaping make the original code barely readable.  Your proposed change looks better :+1:

A few notes to improve this:

1. Regexp are not anchored, so this is equivalent to:
```suggestion
                WHERE array_to_string(acl.defaclacl, ' + ') ~ '(\")?${database_read_only_username}(\")?=r/(\")?${database_username}(\")?'
```

2. The optional quotes work in pairs, so if an opening quote is present, we must match a closing one. We can do this with captures:

```suggestion
                WHERE array_to_string(acl.defaclacl, ' + ') ~ '(\"?)${database_read_only_username}\\1=r/(\"?)${database_username}\\2'
```

3. Since we can't match on array items, joining the items is fine but we should put separators before the first item and after the last so that we can ensure we are matching a complete entry:

```suggestion
                WHERE ' + ' || array_to_string(acl.defaclacl, ' + ') || ' + ' ~ ' + (\"?)${database_read_only_username}\\1=r/(\"?)${database_username}\\2 + '
```

smortex (commented):
It looks like the format depend on the actual role names: `puppetdb-read` is quoted but `puppetdb` is not:
```
puppetdb=# SELECT pg_default_acl.defaclacl::text FROM pg_default_acl;
            defaclacl             
----------------------------------
 {"\"puppetdb-read\"=U/puppetdb"}
 {"\"puppetdb-read\"=X/puppetdb"}
 {"\"puppetdb-read\"=r/puppetdb"}
(3 lignes)
```

I think the `-` in `puppetdb-read` trigger the quoting.  I suggested some changes to make the matching more robust in any circumstances.

smortex:
Out of curiosity, I do not alter the default roles and have not hit this issue. Can you share the configuration you use that trigger the issue?

waipeng:
> Out of curiosity, I do not alter the default roles and have not hit this issue. Can you share the configuration you use that trigger the issue?

We are seeing this on Ubuntu 22.04, Puppet Agent 7.26.0, Puppet Server 7.13.0.

Node hiera is pretty simple
```
puppetdb::listen_address: '0.0.0.0'
puppetdb::postgres_version: '12'
postgresql::server::contrib::package_name: 'postgresql-contrib'
puppetdb::java_args:
  '-Xmx': '2g'
```

A puppet run with -d shows the following:

```
Debug: Executing with uid=postgres gid=postgres: 'psql -d puppetdb -t -X -c "SELECT COUNT(*) FROM (SELECT                                                                                                                                                                                                                                                                                                                                   
                  ns.nspname,                                                                                                                                                                                         
                  acl.defaclobjtype,                                                                                                                                                                                                                                                                                                                                                                                                        
                  acl.defaclacl                                                                                                                                                                                       
                FROM pg_default_acl acl                                                                                                                                                                               
                JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid                                                                                                                                                    
                WHERE acl.defaclacl::text ~ '.*\\\"puppetdb-read\\\"=r/puppetdb\\".*'                                                                                                                                 
                AND nspname = 'public') count"'                                                                                                                                                                       
Error: /Stage[main]/Puppetdb::Database::Postgresql/Puppetdb::Database::Read_only_user[puppetdb-read]/Puppetdb::Database::Default_read_grant[puppetdb grant read permission on new objects from puppetdb to puppetdb-read]/Postgresql_psql[grant default select permission for puppetdb-read]: Could not evaluate: Error evaluating 'unless' clause, returned pid 11837 exit 2: 'sh: 8: Syntax error: Unterminated quoted string             
'    
```

Please let me know if you need more information on replicating this. Thanks for your help!

smortex:
lol, I was forgetting that the unbalanced quote was the reason for this PR :facepalm:.  In our infra, we have `manage_dbserver => false`, that's why we never hit this issue…

The file was introduced with the mismatch in #330. The lack of unit tests seems to allow a broken default config for 5 minor releases in a row… Let's fix the code at least, can you check the suggested regexp changes I proposed and if they work as expected adjust the PR?

manifests/database/default_read_grant.pp
waipeng: The second suggestion works. 

The last suggestion doesn't work. The query returns 0 rows, and as such the ALTER table is run each time puppet runs.

FWIW, I've tried to dump some output from psql so it may be clearer.

```
postgres@puppetdb:~$ psql -d puppetdb -t -X -c "SELECT acl.defaclacl FROM pg_default_acl acl"
 {"\"puppetdb-read\"=U/puppetdb"}
 {"\"puppetdb-read\"=X/puppetdb"}
 {"\"puppetdb-read\"=r/puppetdb"}

postgres@puppetdb:~$ psql -d puppetdb -t -X -c "SELECT array_to_string(acl.defaclacl, ' + ') FROM pg_default_acl acl"
 "puppetdb-read"=U/puppetdb
 "puppetdb-read"=X/puppetdb
 "puppetdb-read"=r/puppetdb
```

waipeng:
I have replied on the suggestions you gave, and added a commit with your change. Thanks for looking at this.

I have taken a look at the tests and #344 was done some time ago and the logs are lost. If you can resurrect it, maybe we can all chip in to get tests working again.

manifests/database/default_read_grant.pp
smortex: Hum, this does not work because `+` has a special meaning in a regexp.

Using another non-special char may be better, such as `@`.

manifests/database/default_read_grant.pp
```diff
@@ -19,7 +19,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=r/${database_username}\\\".*'
+                WHERE array_to_string(acl.defaclacl, ' + ') ~ '(\"?)${database_read_only_username}\\1=r/(\"?)${database_username}\\2'
```
smortex, line 22:
```suggestion
                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=r/(\"?)${database_username}\\2@'
```

manifests/database/default_read_grant.pp
```diff
@@ -36,7 +36,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=U/${database_username}\\\".*'
+                WHERE array_to_string(acl.defaclacl, ' + ') ~ '.*(\")?${database_read_only_username}(\")?=U/(\")?${database_username}(\")?.*'
```
smortex, line 39:
```suggestion
                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=U/(\"?)${database_username}\\2@'
```

manifests/database/default_read_grant.pp
```diff
@@ -53,7 +53,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"${database_read_only_username}\\\\\"=X/${database_username}\\\".*'
+                WHERE array_to_string(acl.defaclacl, ' + ') ~ '.*(\")?${database_read_only_username}(\")?=X/(\")?${database_username}(\")?.*'
```
smortex, line 56:
```suggestion
                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=X/(\"?)${database_username}\\2@'
```

smortex (commented):
Let's give this another try.  Can you try these suggestions?

smortex:
lol

a new release happened

So I should adjust my comment https://github.com/puppetlabs/puppetlabs-puppetdb/pull/371#issuecomment-1742376786

> The file was introduced with the mismatch in #330. The lack of unit tests seems to allow a broken default config for 5 minor releases in a row… Let's fix the code at least, can you check the suggested regexp changes I proposed and if they work as expected adjust the PR?

`s/5 minor releases in a row/6 minor releases in a row/` :upside_down_face:  

Cc @austb.  I feel like you are a bit alone to try to do something with this module, I am working on fixing CI, I'll at-you to be sure you not miss it.

waipeng:
@smortex your suggestions are good. I've added them to the PR.

manifests/database/default_read_grant.pp
smortex: Hey!  I see you replaced the other occurrences of `<SPACE>+<SPACE>` with `@` but not this one.  Was it on purpose of just missing?  I think it's better to use the same construct each time for consistency.

smortex:
Hey @waipeng! I see you replaced `<SPACE>+<SPACE>` with `@` in all fragments but one. Was it on purpose of just missing? I think it's better to use the same construct each time for consistency.

waipeng:
My fault due to inexperience with Github PR.

Do  you have preference on keeping the commits you suggested, or is it OK if I squash them all into one commit and put you as co-author? I am having a hard time tracing through the commits in the current form.

smortex:
Feel free to squash everything ;-)

smortex (approved):
LGTM :100: !

waipeng:
Hi, can this be merged and a new version be cut please?
Right now it is broken in default config which is not very healthy.

alaunay:
Hello, no news on this ? permissions are checked every time without this patch. Thanks !

h0tw1r3:
@waipeng can you rebase on main please?

smortex:
Somebody fixed the test suite :scream:.  We need to update the tests according to the code changes :grin:

@waipeng the "Files changed" tab shows the offending tests (beware, the view lacks some `...` symbols so the diff is not really readable around the failures context), this should help adjusting the test suite to make sure we have the expected code.

Thanks!

h0tw1r3:
@waipeng tests are failing due to unrelated CI changes. Rebase on main again should get past that.

spec/support/unit/shared/database.rb
```diff
@@ -72,7 +72,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"#{with[:database_read_only_username]}\\\\\"=r/#{with[:database_username]}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=r/(\"?)${database_username}\\2@'
```
h0tw1r3, line 75: replace `${database_username}` with `#{with[:database_username]}`

spec/support/unit/shared/database.rb
```diff
@@ -92,7 +92,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"#{with[:database_read_only_username]}\\\\\"=U/#{with[:database_username]}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=U/(\"?)${database_username}\\2@'
```
h0tw1r3, line 95: replace `${database_username}` with `#{with[:database_username]}`

spec/support/unit/shared/database.rb
```diff
@@ -112,7 +112,7 @@
                   acl.defaclacl
                 FROM pg_default_acl acl
                 JOIN pg_namespace ns ON acl.defaclnamespace=ns.oid
-                WHERE acl.defaclacl::text ~ '.*\\\\\"#{with[:database_read_only_username]}\\\\\"=X/#{with[:database_username]}\\\".*'
+                WHERE '@' || array_to_string(acl.defaclacl, '@') || '@' ~ '@(\"?)${database_read_only_username}\\1=X/(\"?)${database_username}\\2@'
```
h0tw1r3, line 115: replace `${database_username}` with `#{with[:database_username]}`

h0tw1r3 (changes requested):

h0tw1r3 (approved):

PR #372, opened 2023-09-13
(BACKWARDS-INCOMPATIBLE) Update firewall manifests to use `jump` instead of `action`

david22swan:
As part of the Firewall module rewrite the functionality of the `action` attribute has been rolled into the `jump` attribute, the two of them both managing the Firewall jump value.

### DIFF
--- a/manifests/server/firewall.pp
+++ b/manifests/server/firewall.pp
@@ -11,17 +11,17 @@ class puppetdb::server::firewall (
 
   if ($open_http_port) {
     firewall { "${http_port} accept - puppetdb":
-      dport  => $http_port,
-      proto  => 'tcp',
-      jump   => 'accept',
+      dport => $http_port,
+      proto => 'tcp',
+      jump  => 'accept',
     }
   }
 
   if ($open_ssl_port) {
     firewall { "${ssl_port} accept - puppetdb":
-      dport  => $ssl_port,
-      proto  => 'tcp',
-      jump   => 'accept',
+      dport => $ssl_port,
+      proto => 'tcp',
+      jump  => 'accept',
     }
   }
 }
--- a/manifests/server/firewall.pp
+++ b/manifests/server/firewall.pp
@@ -13,7 +13,7 @@ class puppetdb::server::firewall (
     firewall { "${http_port} accept - puppetdb":
       dport  => $http_port,
       proto  => 'tcp',
-      action => 'accept',
+      jump   => 'accept',
     }
   }
 
@@ -21,7 +21,7 @@ class puppetdb::server::firewall (
     firewall { "${ssl_port} accept - puppetdb":
       dport  => $ssl_port,
       proto  => 'tcp',
-      action => 'accept',
+      jump   => 'accept',
     }
   }
 }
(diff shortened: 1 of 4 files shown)


manifests/server/firewall.pp
```diff
@@ -11,15 +11,15 @@
     firewall { "${http_port} accept - puppetdb":
       dport  => $http_port,
       proto  => 'tcp',
-      action => 'accept',
+      jump   => 'accept',
```
corporate-gadfly, line 14: May want to reduce the extra space, in case of lint failures.

david22swan: Fixed

joshcooper:
Is the `jump` parameter supported by all versions of the firewall module listed in metadata.json?

david22swan:
`jump` was always present, `action` was created to represent 3 different possible values that can be passed to the iptables jump value but with the release of v7.0.0 `action` has been removed and `jump` is now one value

joshcooper:
> There would likely be errors from this if used alongside versions of firewall less than 7, but is needed for versions greater than 7. 

Should this be bumped to >= 7 and < 8?

https://github.com/puppetlabs/puppetlabs-puppetdb/blob/4a02618679da55c0e7f942164fa3c10645af7de7/metadata.json#L21

david22swan:
Done

david22swan:
> > There would likely be errors from this if used alongside versions of firewall less than 7, but is needed for versions greater than 7.
> 
> Should this be bumped to >= 7 and < 8?
> 
> https://github.com/puppetlabs/puppetlabs-puppetdb/blob/4a02618679da55c0e7f942164fa3c10645af7de7/metadata.json#L21

Updated.
Sorry on the wait, was on PTO friday

corporate-gadfly:
Any thoughts on a release?

david22swan:
@joshcooper Could you take a look

joshcooper (approved):
Seems ok to me, but honestly I have no idea.

smortex (approved):
Looks good to go when we are ready for a major version bump

corporate-gadfly:
> Looks good to go when we are ready for a major version bump

Any estimates on when approximately that might be?

corporate-gadfly:
> Looks good to go when we are ready for a major version bump

Any updates? TY.

rafiqueh:
> Looks good to go when we are ready for a major version bump

Happy New Year. Any updates?

david22swan:
@austb Thoughts on when this could go in?

bastelfreak (approved):
@david22swan looks good to me. maybe correct `BUDFIX`? :D and I think we should keep this as breaking change because it bumps a dependency by a view versions.

david22swan:
Had `BUGFIX` just because I didn't have a ticket for it.
Does this look better?
Or at least more accurate

bastelfreak:
Can you unpin the module in https://github.com/puppetlabs/puppetlabs-puppetdb/blob/main/.fixtures.yml#L22 ? Then tests should/could pass.

bastelfreak (approved):

PR #373, opened 2023-09-29
Revert class parameter removal

austb:
@bastelfreak @ekohl I wanted to open this PR to get some more information from the community and for discussion about possibly reverting or changing removal of the parameter in this PR. We were concerned about the backward incompatibility of this change, since a user specifying the `puppetdb_user` in their manifest would result in a catalog failing to compile. 

Is the primary goal of this change to reduce change events after a puppetdb package upgrade (when the default settings are used)?

We had a few alternate options that we'd like to get feedback on.

1. Revert this change and change the ownership of the config file in the package so that it matches the puppetdb:puppetdb settings in the module
1. Modify this change to allow a user to specify `root` as the owner to avoid the change event on upgrade
1. Leave this change and re-add the parameter with a deprecation note so that users catalogs would continue to compile

### DIFF
--- a/manifests/server/global.pp
+++ b/manifests/server/global.pp
@@ -2,6 +2,7 @@
 class puppetdb::server::global (
   $vardir         = $puppetdb::params::vardir,
   $confdir        = $puppetdb::params::confdir,
+  $puppetdb_user  = $puppetdb::params::puppetdb_user,
   $puppetdb_group = $puppetdb::params::puppetdb_group,
 ) inherits puppetdb::params {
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
I pointed this out a few times: the whole module is in a bad shape. there are zero public tests. I would highly appreciate it if the CAT team takes over maintenance of it (or when a pipeline is added based on their standards).

ekohl:
My reasoning is that the service should not have permissions to modify its own config. It's managed by Puppet and giving it the permission to modify can only break things. So we make `root` the owner. It does need to read the config, so we use the group `puppetdb`. Then nobody else should read it, since the file contains credentials. So mode `0640` is used.

I haven't checked packaging, but that should apply the same permissions.

PR #376, opened 2023-12-14
make puppet_confdir configurable

h0tw1r3:
$settings::<> vars should generally be configurable because they hold the compilers value. This can be a problem if for example you're trying to use this module with Bolt, which returns a temp directory.

I'm not a fan of the globals (or params) class, but this gets the job done in lieu of overhauling the module to meet current standards.

_Hopefully issues can be enabled soon so we can start collaborating on bringing this module into the current century._

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -2,6 +2,7 @@
 class puppetdb::globals (
   $version                      = 'present',
   $database                     = 'postgres',
+  Stdlib::Absolutepath $puppet_confdir = $settings::confdir,
   ) {
 
   if !(fact('os.family') in ['RedHat', 'Suse', 'Archlinux', 'Debian', 'OpenBSD', 'FreeBSD']) {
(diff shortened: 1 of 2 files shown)


manifests/globals.pp
```diff
@@ -2,6 +2,7 @@
 class puppetdb::globals (
   $version                      = 'present',
   $database                     = 'postgres',
+  String $puppet_confdir        = $settings::confdir,
```
smortex, line 5: We can rely on the stdlib datatype for a filename:
```suggestion
  Stdlib::Absolutepath $puppet_confdir = $settings::confdir,
```

smortex (approved):
That look reasonable to me.  Indeed this module need love, and I really would like to see it receive some…

manifests/globals.pp
smortex: Oops, wanted to mark this as a regular comment. Can you fix this @h0tw1r3?

PR #378, opened 2024-01-03
Preamble

spotter-puppet:
Added preamble to README at request of Puppet Community leadership.

PR #379, opened 2024-01-05
Change default postgres version to 14

jonathannewman:
Postgres 11 is no longer supported as an installation method from Postgres.org. As a result, installations will fail using `puppetlabs-postgresql` with failures like:

```
   Error: /Stage[main]/Postgresql::Server::Install/Package[postgresql-server]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install postgresql11-server' returned 1: One of the configured repositories failed (PostgreSQL 11 7 - x86_64),
     and yum doesn't have enough cached data to continue. At this point the only
     safe thing yum can do is fail. There are a few ways to work "fix" this:

         1. Contact the upstream for the repository and get them to fix the problem.

         2. Reconfigure the baseurl/etc. for the repository, to point to a working
            upstream. This is most often useful if you are using a newer
            distribution release than is supported by the repository (and the
            packages for the previous distribution release still work).

         3. Disable the repository, so yum won't use it by default. Yum will then
            just ignore the repository until you permanently enable it again or use
            --enablerepo for temporary usage:

                yum-config-manager --disable yum.postgresql.org

         4. Configure the failing repository to be skipped, if it is unavailable.
            Note that yum will try to contact the repo. when it runs most commands,
            so will have to try and fail each time (and thus. yum will be be much
            slower). If it is a very temporary problem though, this is often a nice
            compromise:

                yum-config-manager --save --setopt=yum.postgresql.org.skip_if_unavailable=true

    failure: repodata/repomd.xml from yum.postgresql.org: [Errno 256] No more mirrors to try.
    https://download.postgresql.org/pub/repos/yum/11/redhat/rhel-7-x86_64/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
```

To make matters more complicated, `puppetlabs-postgresql` is impacted by https://yum.postgresql.org/news/pgdg-rpm-repo-gpg-key-update/

see https://github.com/puppetlabs/puppetlabs-postgresql/issues/1565

so this too will not work until a new version of that is released with a fix.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -24,11 +24,7 @@ class puppetdb::params inherits puppetdb::globals {
     $manage_pg_repo            = false
   }
 
-  if $puppetdb_version in ['latest','present'] or versioncmp($puppetdb_version, '7.0.0') >= 0 {
-    $postgres_version          = '11'
-  } else {
-    $postgres_version          = '9.6'
-  }
+  $postgres_version          = '14'
 
   $puppetdb_major_version = $puppetdb_version ? {
     'latest'  => '8',
--- a/spec/acceptance/standalone_spec.rb
+++ b/spec/acceptance/standalone_spec.rb
@@ -6,8 +6,7 @@ describe 'standalone' do
   let(:puppetdb_params) {}
   let(:puppetdb_master_config_params) {}
 
-  # TODO: temporary work-around for EL installs
-  let(:postgres_version) { "(getvar('facts.os.family') == 'RedHat') ? { true => '12', default => undef }" }
+  let(:postgres_version) { 'undef' } # default
   let(:manage_firewall) { "(getvar('facts.os.family') == 'RedHat' and Integer(getvar('facts.os.release.major')) > 7)" }
 
   describe 'with defaults' do
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -14,9 +14,7 @@ fixtures:
     augeas_core: 'https://github.com/puppetlabs/puppetlabs-augeas_core.git'
     postgresql: 'https://github.com/puppetlabs/puppetlabs-postgresql.git'
     firewall: 'https://github.com/puppetlabs/puppetlabs-firewall.git'
-    postgresql:
-      repo: 'https://github.com/puppetlabs/puppetlabs-postgresql.git'
-      ref: 'v9.2.0'
+    postgresql: 'https://github.com/puppetlabs/puppetlabs-postgresql.git'
     firewall: 'https://github.com/puppetlabs/puppetlabs-firewall.git'
   symlinks:
     puppetdb: '#{source_dir}'
--- a/metadata.json
+++ b/metadata.json
@@ -14,7 +14,7 @@
     },
     {
       "name": "puppetlabs/postgresql",
-      "version_requirement": ">= 6.5.0 < 10.0.0"
+      "version_requirement": ">= 6.5.0 < 11.0.0"
     },
     {
       "name": "puppetlabs/firewall",
--- a/spec/support/acceptance/shared/puppetdb.rb
+++ b/spec/support/acceptance/shared/puppetdb.rb
@@ -3,19 +3,7 @@
 shared_examples 'puppetdb' do
   let(:pp) do
     <<~PP
-    # FIXME: temporary work-around for EL installs
     if $facts['os']['family'] == 'RedHat' {
-      $gpg_key_file = $facts['os']['release']['major'] ? {
-        '7'     => 'PGDG-RPM-GPG-KEY-RHEL7',
-        default => 'PGDG-RPM-GPG-KEY-RHEL',
-      }
-      file { "/etc/pki/rpm-gpg/${gpg_key_file}":
-        source => "https://download.postgresql.org/pub/repos/yum/keys/${gpg_key_file}",
-      }
-      -> Yumrepo <| tag == 'postgresql::repo' |> {
-        gpgkey => "file:///etc/pki/rpm-gpg/${gpg_key_file}",
-      }
-
       # Work-around EL systemd in docker bug affecting forked services
       file_line { 'puppetdb-unit-remove-pidfile':
         path               => '/lib/systemd/system/puppetdb.service',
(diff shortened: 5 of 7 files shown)


manifests/params.pp
```diff
@@ -25,11 +25,7 @@
     $manage_pg_repo            = false
   }
 
-  if $puppetdb_version in ['latest','present'] or versioncmp($puppetdb_version, '7.0.0') >= 0 {
-    $postgres_version          = '11'
-  } else {
-    $postgres_version          = '9.6'
-  }
+  $postgres_version          = '14'
```
smortex, line 28: My guess is that not passing an explicit version by default and letting the puppetlabs-postgresql module default to be used would be more portable.  Unfortunately, this module is currently so broken testing anything require way more efforts that I can afford :unamused:

The fact that the module manage PostgreSQL by default is a bad practice IMHO.  ATM, my recommendation for integrating PuppetDB in a control repo would be to have a puppetdb profile like this (have a dedicated PostgreSQL / firewall profile, disable firewall and dbserver, and use whatever settings you want):

```puppet
class profile::puppetdb {
  include profile::postgresql

  class { 'puppetdb':
    manage_firewall  => false,
    manage_dbserver  => false,
    node_ttl         => '0d',
    node_purge_ttl   => '0d',
  }

  class { 'puppetdb::master::config':
    manage_report_processor => true,
    enable_reports          => true,
  }
}
```

bastelfreak: > this module is currently so broken testing anything require way more efforts that I can afford

I feel so the same. And the lacking CI pipeline is really a problem for me when I review PRs. I suggest to wait for #378 until we merge anything here.

jonathannewman: I can understand that perspective. An alternative one is that updating the value doesn't make things worse than they already are and will make it functional. the "11" value is known to not work and that "14" does. The application has specific expectations about the version of Postgres used and 8.x expects 14. Updating the value doesn't make things worse than they already are.

nabertrand:
Having the official puppetdb module be broken is not a good look for Puppet. Also, why aren't Issues enabled on this repo? Is there a different way for users to report issues?

binford2k:
Issues should be enabled, so I turned them back on.

Per the [Jan 9](https://github.com/voxpupuli/community-triage/wiki#january-9-2024) Vox sync, PMs for our Content and PE teams are looking at the module and discussing alternatives. We'll have more info at the next sync.

corporate-gadfly:
Since PostgreSQL 11 repos have been moved to archived yum repos, I have to tell `puppetdb` and `postgresql` to not manage the yum repos.

My workaround is to use:
```
postgresql::globals::manage_package_repo: false
puppetdb::manage_package_repo: false
```
And then manually set `enabled=0` in `yum.postgresql.org.repo` and create a new `pgdg-11.repo` pointing to `baseurl=https://yum-archive.postgresql.org/11/redhat/rhel-8-x86_64`.

Hope that helps someone.

bastelfreak:
@jonathannewman can you please also unpin the postgresql module in.fixtures.yml and ensure the metadata.json allows the latest major version?

jonathannewman:
> @jonathannewman can you please also unpin the postgresql module in.fixtures.yml and ensure the metadata.json allows the latest major version?

Yep.  Done.

manifests/params.pp
bastelfreak: @jonathannewman stupid question, will this upgrade existing installations from 11 or whatever to 14? We should reflect this properly in the CHANGELOG.md

jonathannewman: Yeah good point.

jonathannewman:
I'm not planning on resolving all those new spec issues.  They existed prior to my PR, and I frankly don't have any more time to devote to this.

bastelfreak:
@jonathannewman that's not correct. We fixed the pipeline and it passes on master. Right now it complains about an outdated REFERENCE.md, because you adjusted the puppet-strings documentation but didn't regenerate it. you can do it by running:
`bundle exec rake strings:generate:reference`

manifests/server/validate_db.pp
```diff
@@ -1,21 +0,0 @@
...
-class puppetdb::server::validate_db (
-  $database_host       = $puppetdb::params::database_host,
-  $database_port       = $puppetdb::params::database_port,
-  $database_username   = $puppetdb::params::database_username,
-  $database_password   = $puppetdb::params::database_password,
-  $database_name       = $puppetdb::params::database_name,
-  $jdbc_ssl_properties = $puppetdb::params::jdbc_ssl_properties,
-) inherits puppetdb::params {
-  if ($database_password != undef and $jdbc_ssl_properties == false) {
-    postgresql::validate_db_connection { 'validate puppetdb postgres connection':
```
bastelfreak, line 13: should we actually delete this? There is https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/lib/puppet/type/postgresql_conn_validator.rb to check if the connection works

h0tw1r3: https://github.com/puppetlabs/puppetlabs-puppetdb/pull/391

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Release notes for the puppetlabs-puppetdb module.
 * Drop support for Puppet 6
 * Drop support for Ubuntu 16.04 and Scientific Linux 7
 * Remove all code related to embedded database
+* Require and install postgres 14 by default
```
bastelfreak, line 9: we don't need to update the changelog here, the release prep task will rewrite it.

spec/acceptance/standalone_spec.rb
```diff
@@ -6,8 +6,7 @@
   let(:puppetdb_params) {}
   let(:puppetdb_master_config_params) {}
 
-  # TODO: temporary work-around for EL installs
-  let(:postgres_version) { "(getvar('facts.os.family') == 'RedHat') ? { true => '12', default => undef }" }
+  let(:postgres_version) { 'undef' } # default
```
bastelfreak, line 9: do we need this let block at all if the value is undef?

h0tw1r3: if the shared example was reworked, no.
the premise is to support testing different versions and/or upgrading in place.

bastelfreak (approved):

CHANGELOG.md
h0tw1r3: will drop the commit after all tests pass

PR #380, opened 2024-01-11
use the correct value for clientcert in pg_hba.conf for Postgresql 12

jhunt-steds:
12 and up, that is.

PuppetDB 8.0.0 requires Postgresql 14 or later, but starting in 12 the syntax for the clientcert option in pg_hba.conf changed. This PR implements the new syntax when a newer version of Postgresql is requested by the user.

### DIFF
--- a/manifests/database/postgresql_ssl_rules.pp
+++ b/manifests/database/postgresql_ssl_rules.pp
@@ -4,7 +4,7 @@
 define puppetdb::database::postgresql_ssl_rules (
   String $database_name,
   String $database_username,
-  String $postgres_version,
+  String[2,3] $postgres_version,
   String $puppetdb_server,
 ) {
   $identity_map_key = "${database_name}-${database_username}-map"
--- a/manifests/database/postgresql_ssl_rules.pp
+++ b/manifests/database/postgresql_ssl_rules.pp
@@ -4,9 +4,15 @@
 define puppetdb::database::postgresql_ssl_rules (
   String $database_name,
   String $database_username,
+  String $postgres_version,
   String $puppetdb_server,
 ) {
   $identity_map_key = "${database_name}-${database_username}-map"
+
+  $clientcert_value = Float($postgres_version) >= 12.0 ? {
+    true    => 'verify-full',
+    false   => '1',
+  }
 
   postgresql::server::pg_hba_rule { "Allow certificate mapped connections to ${database_name} as ${database_username} (ipv4)":
     type        => 'hostssl',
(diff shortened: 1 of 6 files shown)


bastelfreak:
@jhunt-steds thanks for the PR, can you please rebase?

jhunt-steds:
Sorry, I apparently did the rebase in some wrong way that closed the PR for some reason; I'm not really a developer and don't understand git well, but it at least does apply cleanly now I think.

jhunt-steds:
Apologies again, it appears I forgot to run the linter/tests after my rebase yesterday. Everything should be correct now.

h0tw1r3 (approved):

manifests/database/postgresql_ssl_rules.pp
```diff
@@ -4,18 +4,24 @@
 define puppetdb::database::postgresql_ssl_rules (
   String $database_name,
   String $database_username,
+  String $postgres_version,
```
bastelfreak, line 7: Can you make the datatype a bit more strict, at least `String[2]`?

bastelfreak: Oh or we could make it Float? That would make more sense I think?

h0tw1r3: I think that's introducing unrelated changes because technically the module still supports postgresql 9. Another PR to update/fix all of the postgres_version types passing around.

bastelfreak: It might still support 9.6, but nothing earlier, so that's a String with 3 chars or a valid Float. Afterwards postgresql changed their versioning scheme. Now I haven't tested If 11 is a valid Float in puppet or if it has to be 11.0. (if we drop 9.6 we could use Integer as datatype).

jhunt-steds: My decision to set the datatype as `String` was based on it being (implied) as `String` in the rest of the module. The `$postgres_version` variable is available as a parameter with no explicit type in manifests/init.pp, with the default value taken from manifests/params.pp, where it is set (to `'11'` or `'9.6'`) based on the requested version of PuppetDB; given the quotes, the value we're going to be getting is a `String` of some kind. Since all I had to do to solve the clientcert problem was plumb the value taken from init.pp down to manifests/database/postgresql_ssl_params.pp, I wanted to keep changes minimal, so I kept the implied type of `String`.

Whether `$postgres_version` should be `Float` or `Integer` instead of `String` is a worthwhile question IMO but also, I humbly submit, outside the scope of this change, as there would me numerous other changes needed as a result. Changing that would also break backwards compatibility for end users (me, for example) who had previously set it as a `String` value, would it not?  The tests still pass if I change it to `String[2,3]` (to accommodate e.g. '16' or '9.6'), so please let me know if you'd like me to go ahead and push that change.

bastelfreak: I think `String[2,3]` is a good way to move forward.

bastelfreak (approved):

jhunt-steds:
Looks like the failing checks were the result of some network error rather than the changes themselves. Both contain a Ruby stack trace from the first test that has this line:

`/home/runner/work/puppetlabs-puppetdb/puppetlabs-puppetdb/vendor/bundle/ruby/2.7.0/gems/net-ssh-7.2.1/lib/net/ssh/buffered_io.rb:64:in `recv': Connection reset by peer - recvfrom(2) (Errno::ECONNRESET)`

PR #382, opened 2024-01-27
wip ci

h0tw1r3:
* pdk templates updated to 3.0.1
* squashed in #344 
* 100% test unit coverage
* working acceptance tests

aside from syntax, there are **no puppet code changes**.


metadata.json
```diff
@@ -78,11 +78,11 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.10.0 < 9.0.0"
+      "version_requirement": ">= 7.0.0 < 9.0.0"
```
bastelfreak, line 81: This drops a significant amount of major versions. Maybe that can be done in a dedicated PR so it's more present in the changelog?

spec/default_facts.yml
```diff
@@ -2,7 +2,8 @@
 #
 # Facts specified here will override the values provided by rspec-puppet-facts.
 ---
-ipaddress: "172.16.254.254"
-ipaddress6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
+networking:
```
bastelfreak: this breaks a lot of stuff, maybe also the tests in this module (see https://github.com/puppetlabs/pdk-templates/issues/543).  If that's the case we should probably fix it manually.

.sync.yml
```diff
@@ -13,7 +13,11 @@
     - /provision.yaml
 Rakefile:
   default_disabled_lint_checks:
-    - '140chars'
+    - relative
```
bastelfreak, line 18: mhm do we need to disable the `relative` check?

.sync.yml
```diff
@@ -13,7 +13,11 @@
...
+    - parameter_types
+    - parameter_documentation
+    - documentation
```
bastelfreak, line 21: does the documentation check actually exist?

spec/defines/database/read_grant_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    schema:                      'public',
+  },
+}
+
+invalid = {
+  'no params': {},
+}
+
+describe 'puppetdb::database::read_grant' do
+  let(:facts) { on_supported_os.take(1).first[1] }
```
bastelfreak, line 18: Why did you limit the tests to one OS instead of all?

spec/unit/classes/init_spec.rb
```diff
@@ -32,7 +32,10 @@ class { 'postgresql::server':
         end
 
         describe 'manifest' do
-          it { is_expected.to compile.with_all_deps }
+          it {
+            pending('platform support to be removed') if facts[:os]['name'] == 'Ubuntu' && (Puppet::Util::Package.versioncmp(facts[:os]['release']['major'], '18.04') < 0)
```
bastelfreak, line 36: what about dropping Ubuntu 16.04 in another PR stead of marking it as pending?

metadata.json
h0tw1r3: I believe that was in the squashed #344
Is this a blocker? I can't see how it would matter. No reasonable way to test on agents that old.

spec/default_facts.yml
h0tw1r3: file is not in the PR.

.sync.yml
h0tw1r3: not sure why that was there, removed...

.sync.yml
h0tw1r3: confirmed
```
pdk bundle exec puppet-lint --help | grep documentation
pdk (INFO): Using Ruby 3.2.2
pdk (INFO): Using Puppet 8.4.0
        --no-documentation-check     Skip the documentation check.
```

spec/defines/database/read_grant_spec.rb
h0tw1r3: Because that's the way it was :) The change was a simple refactor to remove the `let(:facts)` blocks all over the spec tests.
Many of the classes, which in some cases are just resource wrappers, there is no purpose to testing all platforms.

spec/unit/classes/init_spec.rb
h0tw1r3: I agree another PR is needed, and one is open.
I prefer all tests pass on this PR, which should open up merging the backlog.

metadata.json
bastelfreak: if it matters depends on the CHANGELOG.md quality Puppet wants to have. For Vox Pupuli it's important to document those breaking changes. I suggest to make this change in a dedicated PR and rebase your `good-times` branch against it.

PR #385, opened 2024-01-29
modernize test framework and add ci workflow

h0tw1r3:
* squashed in #344 
* pdk templates updated to 3.0.1
* pin fixtures for modules where latest is not supported yet
* 100% test unit coverage with action workflow

aside from syntax, there are **no puppet code changes**

supersedes #382


.fixtures.yml
```diff
@@ -3,13 +3,29 @@ fixtures:
...
     cron: 'https://github.com/voxpupuli/puppet-cron.git'
-    cron_core: 'https://github.com/puppetlabs/puppetlabs-cron_core.git'
-    yumrepo_core: 'https://github.com/puppetlabs/puppetlabs-yumrepo_core.git'
-    augeas_core: 'https://github.com/puppetlabs/puppetlabs-augeas_core.git'
+    provision: 'https://github.com/puppetlabs/provision.git'
+    puppet_agent: 'https://github.com/puppetlabs/puppetlabs-puppet_agent.git'
+    facts: 'https://github.com/puppetlabs/puppetlabs-facts.git'
+    cron_core:
+      repo: https://github.com/puppetlabs/puppetlabs-cron_core.git
+      puppet_version: ">= 6.0.0"
```
bastelfreak, line 16: We only test on puppet 7 and newer, so we don't need those constraints.

.fixtures.yml
```diff
@@ -3,13 +3,29 @@ fixtures:
...
+  forge_modules:
+    postgresql:
+      repo: 'puppetlabs/postgresql'
+      ref: "9.2.0"
+    firewall:
+      repo: 'puppetlabs/firewall'
+      ref: "6.0.0"
+  symlinks:
+    puppetdb: '#{source_dir}'
```
bastelfreak, line 31: I think the symlink isn't required

.puppet-lint.rc
```diff
@@ -1 +1,5 @@
 --relative
+--no-parameter_types-check
+--no-parameter_documentation-check
+--no-documentation-check
```
bastelfreak, line 4: I am not aware of such a plugin, I need to check if it actually exists. Does puppet-lint find anything if you don't disable this check?

h0tw1r3: confirmed, without `--no-documentation-check`:
```
manifests//database/read_only_user.pp - WARNING: defined type not documented on line 14 (check: documentation)
```

h0tw1r3: each of those checks must be disabled for puppet-lint to pass.

.fixtures.yml
```diff
@@ -1,15 +1,21 @@
...
-    systemd: 'https://github.com/camptocamp/puppet-systemd.git'
-    cron: 'https://github.com/voxpupuli/puppet-cron.git'
+    provision: 'https://github.com/puppetlabs/provision.git'
+    puppet_agent: 'https://github.com/puppetlabs/puppetlabs-puppet_agent.git'
+    facts: 'https://github.com/puppetlabs/puppetlabs-facts.git'
     cron_core: 'https://github.com/puppetlabs/puppetlabs-cron_core.git'
     yumrepo_core: 'https://github.com/puppetlabs/puppetlabs-yumrepo_core.git'
     augeas_core: 'https://github.com/puppetlabs/puppetlabs-augeas_core.git'
+  forge_modules:
+    apt: 'puppetlabs/apt'
```
bastelfreak, line 11: for Vox Pupuli we always reference the git branches. That will pull in changes, for unit tests, before they are released. That allows us to detect breaking changes before they are released. For acceptance tests we pull in the latest forge releases.

.puppet-lint.rc
bastelfreak: lets keep this disabled here. But I need to dig a bit through puppet-lint, I've the feeling parameter and parameter_documentation are a bit redundant :D

spec/acceptance/basic_spec.rb
```diff
@@ -1,17 +1,28 @@
-require 'beaker-puppet'
-require 'beaker-pe'
 require 'spec_helper_acceptance'
 
 describe 'basic tests:' do
   it 'make sure we have copied the module across' do
     # No point diagnosing any more if the module wasn't copied properly
-    shell('ls /etc/puppetlabs/code/modules/puppetdb') do |r|
+    run_shell('ls /etc/puppetlabs/code/environments/production/modules/puppetdb') do |r|
```
bastelfreak, line 6: I doubt that this check brings any value :D

h0tw1r3: i have a whole set of acceptance changes for the next PR. This was part of #344. My next changes remove this test.

.fixtures.yml
h0tw1r3: In principal I disagree. What is being tested should be what is being delivered.

Often I am dealing with the spaghetti of module dependencies in large control repo's. If I see a "tests passing" badge or CI, it's reasonable to expect you can deploy the module and know it's going to work with the dependencies stated in the metadata. Fixtures can't be the source of truth.

If a preview of upcoming changes is needed, that's easy enough to set up a CI job.

h0tw1r3: Perfect example is this module. If I changed the fixtures to git, it would pull in the latest versions, and viola! all the acceptance tests pass. But alas, that's not reality.

.puppet-lint.rc
h0tw1r3: I tested them all independently. I'm sure there's overlap, but something leaked through every time.

.fixtures.yml
h0tw1r3: I really like that idea. Will come up with a PR for running tests with the HEAD of all deps.

There's also the (currently disabled) dependency checker, which is the dummy light of tests.

bastelfreak: > In principal I disagree. What is being tested should be what is being delivered.

that's what the acceptance tests do. The goal of the unit tests is to catch the errors before they are released (also pulling from github is usually faster compared to pulling from the forge).

But as mentioned in my initial comment, that's the strategy we (Vox Pupuli) use for Vox Pupuli. I think it makes sense, but I don't know what kind of strategy Puppet has.

h0tw1r3: @bastelfreak learned something new today. updated commit.

bastelfreak (approved):

PR #388, opened 2024-02-04
Cleanup reference documentation

h0tw1r3:
* moves parameter reference from README with little to no changes to code.
* removes all mention of embedded support.
* adds reference validation task to ci


manifests/globals.pp
```diff
@@ -1,7 +1,14 @@
-# Global configuration class for PuppetDB. See README.md for more details.
+# global configuration class for PuppetDB
```
bastelfreak, line 1: Can you prefix those class descriptions with `# @summary`

bastelfreak (approved):

PR #392, opened 2024-02-16
Fix custom database port support

h0tw1r3:
Very helpful to be able to switch ports on a major version change.

Found while testing upgrading from 7.14.0 to 8.0.0 (to be released) of this module
Testing also revealed, deploying to an existing setup (eg. puppetdb + postgresql 11), postgresql 14 server will be deployed but unable to start (port in use by postgresql 11).

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -84,6 +84,8 @@ class puppetdb::database::postgresql (
   $read_database_password      = $puppetdb::params::read_database_password,
   $read_database_host          = $puppetdb::params::read_database_host
 ) inherits puppetdb::params {
+  $port = scanf($database_port, '%i')[0]
+
   if $manage_server {
     class { 'postgresql::globals':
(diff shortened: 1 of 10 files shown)


manifests/database/postgresql.pp
```diff
@@ -84,6 +84,8 @@
   $read_database_password      = $puppetdb::params::read_database_password,
   $read_database_host          = $puppetdb::params::read_database_host
 ) inherits puppetdb::params {
+  $port = scanf($database_port, '%i')[0]
```
bastelfreak, line 87: I'm not sure if this brings any benefit. This casts it to an int in complicated?

bastelfreak (approved):

manifests/database/postgresql.pp
h0tw1r3: I agree. The type conversion, even if it was `Integer(...)` is fragile.
Made a note to address this a part of a comprehensive "type everything" PR.

PR #398, opened 2024-04-12
replace gha workflows

h0tw1r3:
[cat](https://github.com/puppetlabs/cat-github-actions) workflows appear to be working, and should fix the broken release workflows.


smortex (approved):
I am never sure about what version of the github actions is run but this seems to be the right one, so LGTM!

PR #402, opened 2024-05-02
Fix lower bound the of puppetlabs-postgresql dependency

### LINKED ISSUE #401: Error 'has no parameter named port'
## Describe the Bug
When trying to compile the catalog with puppetdb 8.0.0 and postgresql 9.2.0, it fails with error 'has no parameter named 'port'.

The error is located in file /etc/puppetlabs/code/environments/prod/modules/puppetdb/manifests/database/postgresql.pp at line 149

smortex:
The changes introduced in #392 require changes that where introduced in
version 10.0.0 of the puppetlabs-postgresql module, but the lower
version of the module was not properly updated in this PR.

Fix the lower bound of the dependency to indicate that version 10 of the
puppetlabs-postgresql module is required for proper operation.

Fixes #401

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -14,7 +14,7 @@
     },
     {
       "name": "puppetlabs/postgresql",
-      "version_requirement": ">= 9.2.0 < 11.0.0"
+      "version_requirement": ">= 10.0.0 < 11.0.0"
     },
     {
       "name": "puppetlabs/firewall",


bastelfreak (approved):

smortex:
Labelled this as a bugfix because #392 should have been marked as a breaking-change (but was not) and the last version is a major version bump which is broken under some circumstances that this PR intend to fix.

PR #404, opened 2024-05-03
Add a `puppetdb_version` fact with PuppetDB version

rwaffen:
The code changes add a new Facter fact called `puppetdb_version` that retrieves the version of PuppetDB installed on the system. It uses the `puppetdb --version` command to fetch the version and returns it as a fact value.

### DIFF
--- a/lib/facter/puppetdb_version.rb
+++ b/lib/facter/puppetdb_version.rb
@@ -3,11 +3,6 @@ Facter.add(:puppetdb_version) do
 
   setcode do
     output = Facter::Core::Execution.execute('puppetdb --version')
-
-    if output.nil?
-      nil
-    else
-      output.split(':').last.strip
-    end
+    output.split(':').last.strip
   end
 end
--- a/lib/facter/puppetdb_version.rb
+++ b/lib/facter/puppetdb_version.rb
@@ -1,0 +1,13 @@
+Facter.add(:puppetdb_version) do
+  confine { Facter::Util::Resolution.which('puppetdb') }
+
+  setcode do
+    output = Facter::Core::Execution.execute('puppetdb --version')
+
+    if output.nil?
+      nil
+    else
+      output.split(':').last.strip
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/facter/puppetdb_version.rb
```diff
@@ -0,0 +1,13 @@
+Facter.add(:puppetdb_version) do
+  confine { Facter::Util::Resolution.which('puppetdb') }
+
+  setcode do
+    output = Facter::Core::Execution.execute('puppetdb --version')
+
+    if output.nil?
+      nil
```
bastelfreak, line 8: I think we don't need to return nil since we confine on the puppetdb binary?

rwaffen: thought i add this in case of `puppetdb` returns some error.

rwaffen: removed it.

bastelfreak (approved):

smortex (approved):

Geod24:
Release v8.1.0 is broken for some people because of this.
See issues #412 and #413 . @rwaffen I needed the following to work around the problem on Debian:
```diff
diff --git a/puppet/3rd-party-modules/puppetdb/lib/facter/puppetdb_version.rb b/puppet/3rd-party-modules/puppetdb/lib/facter/puppetdb_version.rb
index b04fb26ce..f75847652 100644
--- a/puppet/3rd-party-modules/puppetdb/lib/facter/puppetdb_version.rb
+++ b/puppet/3rd-party-modules/puppetdb/lib/facter/puppetdb_version.rb
@@ -2,7 +2,7 @@ Facter.add(:puppetdb_version) do
   confine { Facter::Util::Resolution.which('puppetdb') }

   setcode do
-    output = Facter::Core::Execution.execute('puppetdb --version')
-    output.split(':').last.strip
+    output = Facter::Core::Execution.execute('puppetdb version')
+    output.split('=').last.strip
   end
 end
```

smortex:
> ```
> puppetdb version
> ```

On Debian (AIO packages) and FreeBSD (packages built from Puppetlabs tarballs), I get:

```
% puppetdb version
puppetdb: 'version' is not a puppetdb command. See 'puppetdb --help'.
% puppetdb --version
puppetdb version: 8.5.1
```

What did you install?

Issue #386, opened 2024-01-31
Define multiple puppetdb servers in hiera


Szparki:
## Use Case
As far as I know as of now you can only define a single PuppetDB server in hieradata, I have two PuppetDB servers for the same Puppet Master. It would be good to support two or more.

## Describe the Solution You Would Like
support something like the following in hiera:
```
puppet_db:
  server
    - server1
    - server2
    - server N
```
Then in the manifest file we can do the following:
```
$puppet_db = hiera_hash('puppet_db')

  class { 'puppet::server::puppetdb':
      * => $puppet_db 
    }
```
## Describe Alternatives You've Considered
Not managing `/etc/puppetlabs/puppet/puppetdb.conf` with this module

## Additional Context
Maybe there is a solution for this that I couldn't find in the documentation

Szparki:
I was able to do a _workaround_ using the the following in hieradata:
```
puppet_db:
  puppetdb_server: 'server1:8081/,https://server2
  puppetdb_port: 8081
```
And using `puppetdb::master::config` class in manifest:
```
  class { 'puppetdb::master::config':
      * => $puppet_db 
  }
```

alexjfisher:
> I was able to do a _workaround_ using the the following in hieradata:
> 
> ```
> puppet_db:
>   puppetdb_server: 'server1:8081/,https://server2
>   puppetdb_port: 8081
> ```
> 
> And using `puppetdb::master::config` class in manifest:
> 
> ```
>   class { 'puppetdb::master::config':
>       * => $puppet_db 
>   }
> ```

Didn't this break the `puppetdb_conn_validator` type? https://github.com/puppetlabs/puppetlabs-puppetdb/blob/143dc4289ab6f1854e1cb0a8e92f57780ee06235/manifests/master/config.pp#L175

PR #409, opened 2024-05-08
support multiple puppetdb servers

h0tw1r3:
migrated from PR #351 originally submitted by @bassonj

per documentation at https://puppet.com/docs/puppetdb/7/puppetdb_connection.html#server_urls

### DIFF
--- a/manifests/database/ssl_configuration.pp
+++ b/manifests/database/ssl_configuration.pp
@@ -6,7 +6,7 @@ class puppetdb::database::ssl_configuration (
   $database_username           = $puppetdb::params::database_username,
   $read_database_username      = $puppetdb::params::read_database_username,
   $read_database_host          = $puppetdb::params::read_database_host,
-  $puppetdb_server             = $puppetdb::params::puppetdb_server,
+  Variant[String,Array[String, 1]] $puppetdb_server = $puppetdb::params::puppetdb_server,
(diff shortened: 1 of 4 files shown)


h0tw1r3:
I don't really own this per-say, nor have I tested it. The original PR was abandoned, and I was unable to rebase as the author's repo was migrated into another org.

h0tw1r3:
Seems to address #409 but needs work for anyone who wants to pick it up.

h0tw1r3:
Seems to address #386 but needs to be finished. Original author is no longer working on it if someone else is interested in picking this up.

manifests/database/ssl_configuration.pp
```diff
@@ -6,7 +6,7 @@
   $database_username           = $puppetdb::params::database_username,
   $read_database_username      = $puppetdb::params::read_database_username,
   $read_database_host          = $puppetdb::params::read_database_host,
-  $puppetdb_server             = $puppetdb::params::puppetdb_server,
+  Variant[String,Array[String, 1]] $puppetdb_server = $puppetdb::params::puppetdb_server,
```
bastelfreak, line 9: I think that's the least we should do. would even be better to enforce IP/FQDN, but I don't know out of my head what kind of values are allowed here.
```suggestion
  Variant[String[1],Array[String[1], 1]] $puppetdb_server = $puppetdb::params::puppetdb_server,
```

PR #411, opened 2024-05-29
Set data types for puppetdb parameters

### LINKED ISSUE #105: Manifest should validate the passed parameters
I did something stupid:

``` puppet
  class { 'puppetdb':
    disable_ssl    => true,
    listen_port    => 'localhost',
    listen_address => '8000'
  }
```

Of course, 'localhost' isn't a valid port identifier and I don't have any devices called '8000'. Since I messed up, I got this error from PuppetDB:

```
2013-11-28 13:11:18,959 ERROR [main] [puppetlabs.utils] Uncaught exception
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number
        at com.puppetlabs.jetty$plaintext_connector.invoke(jetty.clj:61)
        at com.puppetlabs.jetty$create_server.invoke(jetty.clj:70)
        at ring.adapter.jetty$run_jetty.invoke(jetty.clj:79)
        at com.puppetlabs.jetty$run_jetty$fn__4565.invoke(jetty.clj:96)
        at clojure.core$with_redefs_fn.invoke(core.clj:6751)
        at com.puppetlabs.jetty$run_jetty.invoke(jetty.clj:95)
        at com.puppetlabs.puppetdb.cli.services$_main$fn__12093.invoke(services.clj:512)
        at clojure.core$binding_conveyor_fn$fn__4107.invoke(core.clj:1836)
        at clojure.lang.AFn.call(AFn.java:18)
        at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:334)
        at java.util.concurrent.FutureTask.run(FutureTask.java:166)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1146)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
        at java.lang.Thread.run(Thread.java:679)
2013-11-28 13:11:18,963 INFO  [Thread-4] [cli.services] Shutdown request received; puppetdb exiting.
```

Which could have been avoided if puppet threw an error since I passed a string as a port number. :smile:

chambersmp:
Summary:
Set data types to validate puppetdb params. The data type set was configured based on default values supplied within the [params.pp](https://github.com/puppetlabs/puppetlabs-puppetdb/blob/main/manifests/params.pp) class and alternative values recommended in [PuppetDB docs](https://www.puppet.com/docs/puppetdb/7/configure.html).

The aim is to refine user inputs to reduce risk of misconfiguration.

Related Issues:
resolves #105


manifests/database/postgresql.pp
```diff
@@ -91,7 +91,10 @@
   Boolean $password_sensitive  = false,
   Postgresql::Pg_password_encryption $password_encryption = $puppetdb::params::password_encryption,
 ) inherits puppetdb::params {
-  $port = scanf($database_port, '%i')[0]
+  $port = case $database_port.is_a(String) {
```
bastelfreak, line 94: can't we just set Stdlib::Port::Unprivileged as type?

chambersmp: Happy to adjust to Stdlib::Port::Unprivileged to refine usage (1025 - 65535). I can change the data type from `Variant[Stdlib::Port, Pattern[/\A[0-9]+\Z/]]` to just `Stdlib::Port::Unprivileged` however this may be a breaking change for users who define the ports as a String presently. The params.pp default port values are currently Strings. 
https://github.com/puppetlabs/puppetlabs-puppetdb/blob/db5fcf74ad0f532f0a06e3e181c5de249df2cb65/manifests/params.pp#L36

Alternatively I can set it to `Variant[Stdlib::Port::Unprivileged, Pattern[/\A[0-9]+\Z/]]` to retain support for String values whilst confining Integers to unprivileged values.

Any preference?

h0tw1r3 (approved):

manifests/database/postgresql.pp
chambersmp: The code block is required to avoid a breaking change. If we remove the code block, a user may supply a String value that will be rejected by the date types set on the postgresql::server class and postgresql::server:db defined type. 
The code block converts the `String` value '5432' inherited from params.pp into an Integer 5432. The postgresql::server and defined type postgresql::server:db expect the `port` to be of type Stdlib::Port. They do not accept a String.

chambersmp: It may be worth further restricting the ports from `unprivileged` to just `user` to avoid use of the dynamic port range.

chambersmp: @bastelfreak I've further refined the port params to use `user` range. I've also extended data types to all child classes and defined types. Please review again when able.

manifests/database/default_read_grant.pp
```diff
@@ -2,10 +2,10 @@
 #
 # @api private
 define puppetdb::database::default_read_grant (
-  String $database_name,
-  String $schema,
-  String $database_username,
-  String $database_read_only_username,
+  String[1]              $database_name,
+  String                 $schema,
```
smortex, line 6: String without size is questionable.

Prefer this if an empty string is not valid:
```suggestion
  String[1]              $schema,
```

Or this if an empty string is valid:
```suggestion
  String[0]              $schema,
```

Without an explicit size, we don't know if the lower bound is 0 chars or just not specified.

manifests/database/postgresql_ssl_rules.pp
```diff
@@ -2,10 +2,10 @@
 #
 # @api private
 define puppetdb::database::postgresql_ssl_rules (
-  String $database_name,
-  String $database_username,
+  String      $database_name,
+  String      $database_username,
```
smortex, line 6: Same.  I guess:

```suggestion
  String[1]   $database_name,
  String[1]   $database_username,
```

manifests/database/postgresql_ssl_rules.pp
```diff
@@ -2,10 +2,10 @@
...
   String[2,3] $postgres_version,
-  String $puppetdb_server,
+  String      $puppetdb_server,
```
smortex, line 8: Once more, probably:

```suggestion
  String[1]   $puppetdb_server,
```

manifests/database/read_grant.pp
```diff
@@ -2,9 +2,9 @@
 #
 # @api private
 define puppetdb::database::read_grant (
-  String $database_name,
-  String $schema,
-  String $database_read_only_username,
+  String                 $database_name,
+  String                 $schema,
+  String                 $database_read_only_username,
```
smortex, line 7: Same.

```suggestion
  String[1]              $database_name,
  String[1]              $schema,
  String[1]              $database_read_only_username,
```

manifests/master/config.pp
```diff
@@ -90,35 +90,35 @@
...
+  Boolean                                            $manage_report_processor        = false,
+  Boolean                                            $manage_config                  = true,
+  Boolean                                            $create_puppet_service_resource = true,
+  Boolean                                            $strict_validation              = true,
+  Boolean                                            $enable_reports                 = false,
+  Stdlib::Absolutepath                               $puppet_confdir                 = $puppetdb::params::puppet_confdir,
+  Stdlib::Absolutepath                               $puppet_conf                    = $puppetdb::params::puppet_conf,
+  String[1]                                          $terminus_package               = $puppetdb::params::terminus_package,
+  String[1]                                          $puppet_service_name            = $puppetdb::params::puppet_service_name,
+  Integer                                            $puppetdb_startup_timeout       = $puppetdb::params::puppetdb_startup_timeout,
```
smortex, line 119: A negative value does not make sense.  We can also set a lower bound, but it is admittedly less important than for a String where legacy code used to use `''` for "no value":

```suggestion
  Integer[0]                                         $puppetdb_startup_timeout       = $puppetdb::params::puppetdb_startup_timeout,
```

If you decide to adjust these, there are more but I did not added comments to each of them to avoid flooding with these less important changes.

manifests/server/validate_db.pp
```diff
@@ -2,12 +2,12 @@
...
-  $database_username   = $puppetdb::params::database_username,
-  Variant[String[1], Sensitive[String[1]]] $database_password = $puppetdb::params::database_password,
-  $database_name       = $puppetdb::params::database_name,
-  $jdbc_ssl_properties = $puppetdb::params::jdbc_ssl_properties,
+  Stdlib::Host                                       $database_host       = $puppetdb::params::database_host,
+  Variant[Stdlib::Port::User, Pattern[/\A[0-9]+\Z/]] $database_port       = $puppetdb::params::database_port,
+  String[1]                                          $database_username   = $puppetdb::params::database_username,
+  Variant[String[1], Sensitive[String[1]]]           $database_password   = $puppetdb::params::database_password,
+  String[1]                                          $database_name       = $puppetdb::params::database_name,
+  Variant[String, Boolean]                           $jdbc_ssl_properties = $puppetdb::params::jdbc_ssl_properties,
```
smortex, line 10: This is interesting.  The default value in `params.pp` seems to be `''`, but we also compare it to `false` so maybe this can be something like:

```suggestion
  Variant[String[0], Boolean[false]]                 $jdbc_ssl_properties = $puppetdb::params::jdbc_ssl_properties,
```

Which should probably be simplified to something like this in another PR because it breaks backwards compatibility:
```suggestion
  Optional[String[1]]                                $jdbc_ssl_properties = $puppetdb::params::jdbc_ssl_properties,
```

manifests/server/validate_read_db.pp
```diff
@@ -2,12 +2,12 @@
...
-  $database_username   = $puppetdb::params::database_username,
-  Variant[String[1], Sensitive[String[1]]] $database_password = $puppetdb::params::database_password,
-  $database_name       = $puppetdb::params::database_name,
-  $jdbc_ssl_properties = $puppetdb::params::jdbc_ssl_properties,
+  Stdlib::Host                                       $database_host       = $puppetdb::params::database_host,
+  Variant[Stdlib::Port::User, Pattern[/\A[0-9]+\Z/]] $database_port       = $puppetdb::params::database_port,
+  String[1]                                          $database_username   = $puppetdb::params::database_username,
+  Variant[String[1], Sensitive[String[1]]]           $database_password   = $puppetdb::params::database_password,
+  String[1]                                          $database_name       = $puppetdb::params::database_name,
+  Variant[String, Boolean]                           $jdbc_ssl_properties = $puppetdb::params::jdbc_ssl_properties,
```
smortex, line 10: Same

manifests/database/default_read_grant.pp
chambersmp: Thanks @smortex for the suggestion, I've set lower bounds for both strings and integers in the latest commit. I've checked each one for usage of empty strings '' and to guard against negative integers.

manifests/master/config.pp
chambersmp: Each integer has had lower bounds set.

manifests/server/validate_db.pp
chambersmp: As recommended I've set the lower bounds to `String[0]` to allow an empty value `''` and the `Boolean` type to allow only `false`. I agree, it would be worth revisiting in another PR to refactor the param to optional and perhaps to use `Pattern` to validate the value starts with `?` as an expected value would be `?ssl=true`.

h0tw1r3:
FWIW, I would remove the maintenance label so the change *does* show up in the change log and release notes.

manifests/init.pp
```diff
@@ -346,89 +346,89 @@
...
+  Stdlib::Host                                                         $listen_address                    = $puppetdb::params::listen_address,
+  Variant[Stdlib::Port::User, Pattern[/\A[0-9]+\Z/]]                   $listen_port                       = $puppetdb::params::listen_port,
+  Boolean                                                              $disable_cleartext                 = $puppetdb::params::disable_cleartext,
+  Boolean                                                              $open_listen_port                  = $puppetdb::params::open_listen_port,
+  Stdlib::Host                                                         $ssl_listen_address                = $puppetdb::params::ssl_listen_address,
+  Variant[Stdlib::Port::User, Pattern[/\A[0-9]+\Z/]]                   $ssl_listen_port                   = $puppetdb::params::ssl_listen_port,
+  Boolean                                                              $disable_ssl                       = $puppetdb::params::disable_ssl,
+  Optional[Boolean]                                                    $open_ssl_listen_port              = $puppetdb::params::open_ssl_listen_port,
```
bastelfreak, line 356: I think instead of `Optional[Boolean]`, the value in params.pp should be changed to `false` and then use `Boolean` as type.

chambersmp: agreed

chambersmp:
As open_ssl_listen_port is no longer `optional` both params.pp and the unit test shared-examples have been updated to default to `false` instead of `undef`|`nil`.

h0tw1r3 (approved):

smortex (approved):

techsk8:
is there any blocker in getting this PR merged now that's approved?

chambersmp:
> is there any blocker in getting this PR merged now that's approved?

Should be good to go, I haven't had a chance to test on live node so I haven't merged. Passing all unit tests.

zaben903:
Tested changes with a development environment Puppetserver.
Confirmed invalid data such as the example from #105 will return a compile time error.
```puppet
class { 'puppetdb':
  listen_port => 'localhost',
 }
```
```
$ puppet agent -t
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Notice: Requesting catalog from puppet:8140 (127.0.0.1)
Notice: Catalog compiled by puppet-testing
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Puppetdb]: parameter 'listen_port' expects a value of type Stdlib::Port::User = Integer[1024, 49151] or Pattern[/\A[0-9]+\Z/], got String (file: /etc/puppet/code/environments/production/manifests/site.pp, line: 32, column: 3) on node puppet-testing
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

When data is valid, this module works as expected compiling and implementing changes.
Confirmed working with Strings and Integers with pattern matching (where used).

```puppet
class { 'puppetdb':
  ssl_listen_port => 8000, # or '8000'
}

class { 'puppetdb::master::config':
  puppetdb_port => 8000, # or '8000'
}
```
```
puppet agent -t
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Notice: Requesting catalog from puppet:8140 (127.0.0.1)
Notice: Catalog compiled by puppet-testing
Info: Caching catalog for puppet-testing
Info: Applying configuration version '1760595330'
Notice: /Stage[main]/Puppetdb::Server::Jetty/Ini_setting[puppetdb_sslport]/value: value changed 8081 to 8000
Info: Class[Puppetdb::Server::Jetty]: Scheduling refresh of Service[puppetdb]
Notice: /Stage[main]/Puppetdb::Server/Service[puppetdb]: Triggered 'refresh' from 1 event
Notice: /Stage[main]/Puppetdb::Master::Puppetdb_conf/Ini_setting[puppetdbserver_urls]/value: value changed https://puppet-testing:8081/ to https://puppet-testing:8000/
Info: Class[Puppetdb::Master::Puppetdb_conf]: Scheduling refresh of Service[puppetserver]
Notice: /Stage[main]/Puppetdb::Master::Config/Service[puppetserver]: Triggered 'refresh' from 1 event
Notice: Applied catalog in 43.43 seconds
```

PR #415, opened 2024-06-10
Only include 'firewall' module when necessary

Geod24:
If the user is not using an open SSL port or an open HTTP port, we don't need to include this module.


kenyon (approved):

smortex (approved):
Looks reasonable!

Issue #413, opened 2024-06-06
Error while resolving custom fact fact='puppetdb_version'


damluk:
Debian 12, Puppetserver 7.9.5, PuppetDB 7.12.1.

```
Error: Facter: Error while resolving custom fact fact='puppetdb_version', resolution='<anonymous>': undefined method `strip' for nil:NilClass

    output.split(':').last.strip
```

```
setcode do
  output = Facter::Core::Execution.execute('puppetdb --version')
  output.split(':').last.strip                                                                                                                                                                                                                                                                      
end
```

```
$ /usr/sbin/puppetdb --version
Available subcommands:
  version                 Display version information
  services                Run PuppetDB
  upgrade                 Upgrade to latest version and exit
  benchmark               Run development-only benchmarking tool
  fact-storage-benchmark
  help                    Display usage summary
For help on a given subcommand, invoke it with -h

$ /usr/sbin/puppetdb version
version=7.12.1
target_schema_version=80
```

Geod24:
Introduced in https://github.com/puppetlabs/puppetlabs-puppetdb/pull/404
Reverting to v8.0.1 will work around the issue, otherwise look at the comment I posted in the PR.

smortex:
> ```
> $ /usr/sbin/puppetdb --version
> ```

That's not a path I would expect on Debian…

What/How did you install? You seem to be using PuppetServer 7 / PuppetDB 7 on Debian 12 for which Puppet does not provide AIO packages.  I think you can cast your voice here: https://github.com/puppetlabs/community/discussions/65

Regarding the root cause, the version string comes from the puppetdb wrapper generated by ezbake (this code seems to be 10 years old):

https://github.com/puppetlabs/ezbake/blob/5eec28239e177f14a3a6eaafc305119c6b375dfb/resources/puppetlabs/lein-ezbake/staging-templates/cli-app.erb#L60

FYI, the FreeBSD packages are built using the tarballs from https://downloads.puppetlabs.com/puppetdb/ and don't have this issue.  Either there was an issue in older releases that got fixed, or the installation method you use give you a weird `puppetdb` script.

damluk:
> Regarding the root cause, the version string comes from the puppetdb wrapper generated by ezbake (this code seems to be 10 years old):
> 
[...]

PR #416, opened 2024-06-14
Extend fact to not fail on debian packages

rwaffen:
i tried to get around the debian package differences and report a version also if this package is used. sadly somehow i didnt get `puppetdb version` running on my ubuntu testbox. so i tried to get the version from the package itself if it is not the puppetlabs one.

fixes: #413

### DIFF
--- a/lib/facter/puppetdb_version.rb
+++ b/lib/facter/puppetdb_version.rb
@@ -5,7 +5,7 @@ Facter.add(:puppetdb_version) do
     require 'open3'
 
     # check if os is debian/ubuntu and the package is not from puppetlabs
-    if Facter.value(:osfamily) == 'Debian'
+    if Facter.value(:os)('family') == 'Debian'
       package_maintainer = Facter::Core::Execution.execute('apt-cache show puppetdb | grep "Maintainer:" | head -1')
       unless package_maintainer.include? 'Puppet Labs'
         output, status = Open3.capture2('dpkg-query --showformat=\'${Version}\' --show puppetdb')
--- a/lib/facter/puppetdb_version.rb
+++ b/lib/facter/puppetdb_version.rb
@@ -2,26 +2,25 @@ Facter.add(:puppetdb_version) do
   confine { Facter::Util::Resolution.which('puppetdb') }
 
   setcode do
-    require 'open3'
+    command = 'puppetdb --version'
+    splitter = ':'
+    postion = 'last'
 
-    # check if os is debian/ubuntu and the package is not from puppetlabs
-    if Facter.value(:os)('family') == 'Debian'
+    if Facter.value(:os)['family'] == 'Debian'
       package_maintainer = Facter::Core::Execution.execute('apt-cache show puppetdb | grep "Maintainer:" | head -1')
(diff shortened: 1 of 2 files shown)


lib/facter/puppetdb_version.rb
```diff
@@ -2,7 +2,26 @@
...
   setcode do
-    output = Facter::Core::Execution.execute('puppetdb --version')
-    output.split(':').last.strip
+    require 'open3'
+
+    # check if os is debian/ubuntu and the package is not from puppetlabs
+    if Facter.value(:osfamily) == 'Debian'
+      package_maintainer = Facter::Core::Execution.execute('apt-cache show puppetdb | grep "Maintainer:" | head -1')
+      unless package_maintainer.include? 'Puppet Labs'
+        output, status = Open3.capture2('dpkg-query --showformat=\'${Version}\' --show puppetdb')
```
bastelfreak, line 11: is there a reason for using `Open3` here instead of `Facter::Core::Execution.execute`?

rwaffen: so i can do `status.success?` and react if it does not work.

lib/facter/puppetdb_version.rb
```diff
@@ -2,7 +2,25 @@
...
 
   setcode do
-    output = Facter::Core::Execution.execute('puppetdb --version')
-    output.split(':').last.strip
+    command = 'puppetdb --version'
+    splitter = ':'
+    postion = 'last'
+
+    if Facter.value(:os)['family'] == 'Debian'
+      package_maintainer = Facter::Core::Execution.execute('apt-cache show puppetdb | grep "Maintainer:" | head -1')
```
smortex, line 10: Won't this list all *available* versions of the `puppetdb` package (that is the Debian *and* the Puppetlabs ones) regardless of which version is currently installed (if any)?  I feel like taking the first one will not do what we want 100% of the time.

I don't have a Debian 12 system with PuppetDB at hand to check the output ordering, but on Debian 11 where Puppetlabs packages are available `apt-cache show puppetdb` currently list 10 packages from Puppetlabs, and I don't see something that indicate which one is installed…

I feel like `dpkg -s puppetdb` would provide a similar output but *only for the installed* package.  There filtering on the maintainer as you tdo seems to be the best we can do I suppose :shrug:.

smortex (commented):
I really don't like this: it really stands as a packaging issue, and with the lifecycles of Debian releases we will have to carry this cruft for years…  :sob: 

I added in-line notes, I think it is not as reliable as expected as it is now, but not being able to test this because of the missing Debian 12 packages from Puppetlabs do not help :sob:.

rwaffen:
i'll drop my efforts on solving this. its hard to get a testing environment for me. and the packages are completly not like the puppetlabs ones. this is a puppetlabs repo, so we handle puppetlabs packages with puppetlabs commands. if commands fail, report nil and exit.

bastelfreak:
I think this is valid, if someone cares about packages for other distributions they can add support for it.

bastelfreak (approved):

PR #417, opened 2024-06-17
Update CODEOWNERS

chambersmp:
Simple PR, remove code owner that no longer has access

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @puppetlabs/puppetdb @bastelfreak @smortex @h0tw1r3
+* @puppetlabs/puppetdb @bastelfreak @smortex


CODEOWNERS
```diff
@@ -1 +1 @@
-* @puppetlabs/puppetdb @bastelfreak @smortex @h0tw1r3
```
bastelfreak, line 1: @chambersmp why do you remove @h0tw1r3 ?

chambersmp: @bastelfreak @h0tw1r3 doesn't appear to have `write` access anymore to be a Codeowner.

bastelfreak (approved):

bastelfreak:
I'm going to merge this. Because @h0tw1r3 has no access anymore the CODEOWNERS file is invalid which breaks permissions.

Issue #412, opened 2024-06-01
puppetdb fails to start after switching to v8.1.0


vchepkov:
## Describe the Bug
Upgrading to module v8.1.0 renders puppetdb unoperational , service doesn't start
The following log entries are observed in puppetdb.log 
```
2024-06-01T14:24:13.865Z INFO  [p.p.c.services] PuppetDB version 8.5.1
2024-06-01T14:24:13.867Z INFO  [p.p.c.services] Ensuring default database is up to date
2024-06-01T14:24:13.869Z INFO  [c.z.h.HikariDataSource] PDBMigrationsPool: default - Starting...
2024-06-01T14:24:13.883Z INFO  [c.z.h.HikariDataSource] PDBMigrationsPool: default - Start completed.
2024-06-01T14:24:16.900Z ERROR [p.p.c.services] Will retry database connection after temporary failure: java.sql.SQLTransientConnectionException: PDBMigrationsPool: default - Connection is not available, request timed out after 3014ms.
2024-06-01T14:24:19.900Z ERROR [p.p.c.services] Will retry database connection after temporary failure: java.sql.SQLTransientConnectionException: PDBMigrationsPool: default - Connection is not available, request timed out after 3000ms.
```

Reverting back to v8.0.1 restores functionality

## Environment
```
postgresql14-libs-14.12-1PGDG.rhel9.x86_64
postgresql14-14.12-1PGDG.rhel9.x86_64
postgresql14-server-14.12-1PGDG.rhel9.x86_64
postgresql14-contrib-14.12-1PGDG.rhel9.x86_64
puppet8-release-1.0.0-8.el9.noarch
puppet-agent-8.6.0-1.el9.x86_64
puppetdb-termini-8.5.1-1.el9.noarch
puppetserver-8.6.0-1.el9.noarch
puppetdb-8.5.1-1.el9.noarch
```

AlmaLinux 9.4

Geod24:
Duplicate of #413

smortex:
@vchepkov If the service fail to restart, there are corrective changes right?  What are they?

Also the message indicate that the server cannot reach the database in 3s and timeout.  Have you tried to measure how long it take to connect?  Maybe some changes delay the connection and a raising the value way workaround the issue?

vchepkov:
That's a good question.  I only used it for bootstrap and it didn't start. I will try to see, what correction v8.0.1 will do

vchepkov:
[...]

PR #421, opened 2024-12-13
fix: Use C.UTF8 locale to create puppetdb database instead of en_US

saimonn:
This should fix the [breaking change](https://github.com/puppetlabs/puppetlabs-puppetdb/commit/0c5c2708f4683a6056ab8bb9e9cae092f09786d1#commitcomment-143826184) introduced by https://github.com/puppetlabs/puppetlabs-puppetdb/commit/0c5c2708f4683a6056ab8bb9e9cae092f09786d1 were the puppetdb can't be created on a system where the locale en_US.UTF-8 is not installed.

This has been also reported in in https://github.com/puppetlabs/puppetlabs-puppetdb/issues/412#issuecomment-2207551604

C.UTF-8 seems a better choice in this case, I'm not a locales expert enough to say if it's the perfect choice.

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -150,7 +150,7 @@ class puppetdb::database::postgresql (
       user     => $database_username,
       password => $database_password,
       encoding => 'UTF8',
-      locale   => 'en_US.UTF-8',
+      locale   => 'C.UTF-8',
       grant    => 'all',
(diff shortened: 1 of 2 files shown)


saimonn:
I'm closing this PR in favor of https://github.com/puppetlabs/puppetlabs-puppetdb/pull/422

PR #422, opened 2024-12-13
feat: use variable for postgresql locale

JGodin-C2C:
This should fix the [breaking change](https://github.com/puppetlabs/puppetlabs-puppetdb/commit/0c5c2708f4683a6056ab8bb9e9cae092f09786d1#commitcomment-143826184) introduced by https://github.com/puppetlabs/puppetlabs-puppetdb/commit/0c5c2708f4683a6056ab8bb9e9cae092f09786d1 were the puppetdb can't be created on a system where the locale EN_us.UTF-8 is not installed.

This has been also reported in in https://github.com/puppetlabs/puppetlabs-puppetdb/issues/412#issuecomment-2207551604

C.UTF-8 seems a better choice in this case, I'm not a locales expert enough to say if it's the perfect choice.

### DIFF
--- a/spec/unit/classes/database/postgresql_spec.rb
+++ b/spec/unit/classes/database/postgresql_spec.rb
@@ -63,7 +63,7 @@ describe 'puppetdb::database::postgresql', type: :class do
             grant:    'all',
             port:     params[:database_port].to_i,
             encoding: 'UTF8',
-            locale:   'en_US.UTF-8',
+            locale:   params[:postgresql_locale],
           )
       }
 
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -49,6 +49,8 @@
 #
 # @param postgresql_ssl_ca_cert_path
 #   Path to the Postgresql SSL CA.
+# @param postgresql_locale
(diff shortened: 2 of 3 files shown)


JGodin-C2C:
/me gently waving in the general direction of  @bastelfreak @smortex ,

bastelfreak:
Hi,
I'm currently holding off reviews until https://www.puppet.com/blog/open-source-puppet-updates-2025 is sorted out.

spec/unit/classes/database/postgresql_spec.rb
```diff
@@ -63,7 +63,7 @@
             grant:    'all',
             port:     params[:database_port].to_i,
             encoding: 'UTF8',
-            locale:   'en_US.UTF-8',
+            locale:   params[:postgresql_locale],
```
smortex, line 66: `postgresql_locale` does not appear in the `params` definition, so I would expect CI to fail, let's see…

zaben903 (changes requested):
LGTM, PR needs to be rebased before it can be merged, @JGodin-C2C once rebased we can get this merged.

PR #424, opened 2025-02-25
add possibility to manage puppetdb terminus package

techsk8:
I'd like to be able to install `openvoxdb-termini` through the main class and currently that's not possible.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -156,6 +156,9 @@
 #   service doesn't start on boot either. Valid values are `true`, `running`,
 #   `false`, and `stopped`.
 #
+# @param puppetdb_terminus_package
+#   The PuppetDB terminus package name in the package manager. Defaults to `puppetdb-termini` or `puppetdb-terminus`.
+#
 # @param puppetdb_user
 #   Puppetdb service user
 #
@@ -396,6 +399,7 @@ class puppetdb (
   $puppetdb_package                        = $puppetdb::params::puppetdb_package,
   $puppetdb_service                        = $puppetdb::params::puppetdb_service,
   $puppetdb_service_status                 = $puppetdb::params::puppetdb_service_status,
+  String[1] $puppetdb_terminus_package     = $puppetdb::params::terminus_package,
   $puppetdb_user                           = $puppetdb::params::puppetdb_user,
   $puppetdb_group                          = $puppetdb::params::puppetdb_group,
   $puppetdb_server                         = $puppetdb::params::puppetdb_server,
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -104,6 +104,7 @@ The following parameters are available in the `puppetdb` class:
 * [`puppetdb_package`](#-puppetdb--puppetdb_package)
 * [`puppetdb_service`](#-puppetdb--puppetdb_service)
 * [`puppetdb_service_status`](#-puppetdb--puppetdb_service_status)
+* [`puppetdb_terminus_package`](#-puppetdb--puppetdb_terminus_package)
 * [`puppetdb_user`](#-puppetdb--puppetdb_user)
 * [`puppetdb_group`](#-puppetdb--puppetdb_group)
 * [`puppetdb_server`](#-puppetdb--puppetdb_server)
@@ -503,6 +504,14 @@ service doesn't start on boot either. Valid values are `true`, `running`,
 `false`, and `stopped`.
 
 Default value: `$puppetdb::params::puppetdb_service_status`
+
+##### <a name="-puppetdb--puppetdb_terminus_package"></a>`puppetdb_terminus_package`
+
+Data type: `String[1]`
+
+The PuppetDB terminus package name in the package manager. Defaults to `puppetdb-termini` or `puppetdb-terminus`.
+
+Default value: `$puppetdb::params::terminus_package`
 
 ##### <a name="-puppetdb--puppetdb_user"></a>`puppetdb_user`
 


manifests/init.pp
```diff
@@ -430,6 +434,8 @@
   Postgresql::Pg_password_encryption $postgresql_password_encryption = $puppetdb::params::password_encryption,
   Optional[Stdlib::Absolutepath] $java_bin = $puppetdb::params::java_bin,
 ) inherits puppetdb::params {
+  include postgresql
```
bastelfreak: why do you include postgresql here now?

manifests/init.pp
```diff
@@ -396,6 +399,7 @@
   $puppetdb_package                        = $puppetdb::params::puppetdb_package,
   $puppetdb_service                        = $puppetdb::params::puppetdb_service,
   $puppetdb_service_status                 = $puppetdb::params::puppetdb_service_status,
+  $puppetdb_terminus_package               = $puppetdb::params::puppetdb_terminus_package,
```
bastelfreak: please add a datatype to the parameter
```suggestion
  String[1] $puppetdb_terminus_package     = $puppetdb::params::puppetdb_terminus_package,
```

bastelfreak:
greetings to sys11 from betadots :)

manifests/init.pp
techsk8: I was just testing something temporary on this branch, it's gone now, I should've used some other branch. 🤦🏽‍♂️

manifests/init.pp
techsk8: sure!

KoenDierckx:
Also looking forward to openvox support from this module

jadestorm:
Hi @techsk8 -- so as is this doesn't really "do" anything.  Meaning it adds a parameter but it does not change the code that references $puppetdb::params::terminus_package directly.

However -- you can accomplish this already "as is" with:
puppetdb::master::config::terminus_package: 'openvoxdb-termini'

At least as far as I have tested so far.

techsk8:
Hi @jadestorm,
Well, that depends on how you actually use the `puppetdb` module. I use it contained inside a wrapper si I'm only calling `puppetdb` and passing the params like shown below. So in my case the above is useful and does "do" something:

```puppet
class mywrapper_puppet::puppetdb::upstream (
) {
  assert_private()

  ##
  ## Upstream module
  ##
  class { 'puppetdb':
    certificate_whitelist     => $mywrapper_puppet::puppetdb::certificate_whitelist,
    cipher_suites             => $mywrapper_puppet::puppetdb::cipher_suites,
    command_threads           => $mywrapper_puppet::puppetdb::command_threads,
    database_password         => $mywrapper_puppet::puppetdb::database_password,
    disable_cleartext         => $mywrapper_puppet::puppetdb::disable_cleartext,
    disable_update_checking   => $mywrapper_puppet::puppetdb::disable_update_checking,
    java_args                 => $mywrapper_puppet::puppetdb::java_args,
    manage_firewall           => $mywrapper_puppet::puppetdb::manage_firewall,
    manage_package_repo       => false,
    max_threads               => $mywrapper_puppet::puppetdb::max_threads,
    node_purge_ttl            => $mywrapper_puppet::puppetdb::node_purge_ttl,
    node_ttl                  => $mywrapper_puppet::puppetdb::node_ttl,
    postgres_version          => $mywrapper_puppet::puppetdb::postgres_version,
    puppetdb_terminus_package => $mywrapper_puppet::puppetdb_terminus_package,
    puppetdb_package          => $mywrapper_puppet::puppetdb_package,
    report_ttl                => $mywrapper_puppet::puppetdb::report_ttl,
    ssl_listen_address        => $mywrapper_puppet::puppetdb::ssl_listen_address,
  }
.
.
.
}
```

jadestorm:
Hi @techsk8 !  Sorry yesterday I was in a hurry I meant to show you what I was referring to.  So what I see under files changed is that you added an option to pass in the parameter:
  `String[1] $puppetdb_terminus_package     = $puppetdb::params::terminus_package,`

And some documentation.  However, the part of the code that actually creates the terminus package resource is in:
  https://github.com/puppetlabs/puppetlabs-puppetdb/blob/main/manifests/master/config.pp

Specifically:
```
  package { $terminus_package:
  }
```

Which is set from puppetdb::master:config's parameter:
  `$terminus_package            = $puppetdb::params::terminus_package,`

Now as you can see -- unless you explicitly set that parameter on puppetdb::master::config, it defaults to $puppetdb::params::terminus_package, which would bypass your puppetdb::puppetdb_terminus_package setting.

Now, as is, puppetdb::master::config inherits puppetdb::params, do just changing the default on that class isn't going to accomplish anything.

However, the documentation suggests using it like:
```
node <hostname> {
  # Configure puppetdb and its underlying database
  class { 'puppetdb': }
  
  # Configure the Puppet master to use puppetdb
  class { 'puppetdb::master::config': }
}
```

In which case, you could simply pass the terminus_package param to your call of puppetdb::master::config there.

FWIW I did try your fork and it didn't change what package it was trying to pull in, so I went digging to try to figure out why and that's when I found I could work around it.

techsk8:
I understand what you're saying.
But the way I'm applying puppet code on my nodes is only through hiera config. I'm not using `manifests/site.pp` for anything, but assigning classes and roles, something in the lines of:
```
lookup('classes', { merge => 'unique' }).include
```

Actually, I think the main point of Hiera and Puppet is that it enables you to separate configuration data from your Puppet manifests, making your code more reusable and easier to manage.
But if you prefer managing your node's config with your nodes from `manifests/site.pp` then your point makes sense. 😄
In my wrapper class that I use for managing puppet, in the case of puppetdb the only class I call for that is `puppetdb`and `puppetdb::globals` nothing else, so for me it makes sense to pass in the `$terminus_package` param to `puppetdb` directly.

PR #426, opened 2025-04-08
Fix documentation for PostgreSQL version

yakatz:
The default was changed in #379


zaben903:
Resolved as part of https://github.com/puppetlabs/puppetlabs-puppetdb/pull/436

PR #428, opened 2025-07-17
feature: add possiblity to manage puppetdb terminus

techsk8:
replaces #424

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -104,6 +104,7 @@ The following parameters are available in the `puppetdb` class:
 * [`puppetdb_package`](#-puppetdb--puppetdb_package)
 * [`puppetdb_service`](#-puppetdb--puppetdb_service)
 * [`puppetdb_service_status`](#-puppetdb--puppetdb_service_status)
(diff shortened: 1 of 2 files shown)


zaben903:
@techsk8 functionally this PR appears to be the same as #424. Due to this it encounters the same issues raised by https://github.com/puppetlabs/puppetlabs-puppetdb/pull/424#issuecomment-2787120271 and https://github.com/puppetlabs/puppetlabs-puppetdb/pull/424#issuecomment-2789821055.

By adding the parameter you _could_ manage it through the `puppetdb` class, but this class doesn't handle the terminus package installation, instead `puppetdb::master::config` is used to install the package using the `$terminus_package` variable directly on `puppetdb::master::config`.

```puppet
# @param terminus_package
#   Name of the package to use that represents the PuppetDB terminus code. Defaults
#   to `puppetdb-termini`, when `puppetdb_version` is set to `<= 2.3.x` the default
#   changes to `puppetdb-terminus`.
...
class puppetdb::master::config (
...
  String[1] $terminus_package = $puppetdb::params::terminus_package,
...
package { $terminus_package:
}
```

As your PR is written today, there will be no functional impact.

